PR #9, opened 2024-11-04
[WIP] Add ADR proposal and LSWT draft

mducle:
Adds some developer documentation:

* A proposal to record architectural design decisions in the documentation in the repository
* A draft (WIP) of Linear Spin Wave Theory with reference to the (Matlab) SpinW code.


docs/developers/design/001_linear_spinwave_theory.md
```diff
@@ -0,0 +1,95 @@
...
+Consider a single atom with total spin quantum number $S$, which can be in states labelled by $`S_z=-S,-S+1,...,S-1,S`$.
+In the ordered state, this spin is in the state with maximum $`S_z=S`$.
+A small deviation from this (i.e. initiating a spin wave) will change the state to $`S_z=S-1`$.
+This can be described by the spin lowering operator $`\hat{S}^-`$.
+Likewise, restoring the ordered state from this deviation is described by the spin raising operator $`\hat{S}^+`$.
+
+Now, the _Holstein-Primakoff_ transformation is a mapping between these lowering and raising operators
+to bosonic creation $`\hat{b}^{\dagger}`$ and annihilation $`\hat{b}`$ operators as follows (we drop the hats):
+
+```math
```
lucas-wilkins: Something that's been a real blocker trying to understand all this is how you can do a square root on these objects. I guess if they're matrices you can do a Cholesky decomposition, but later you talk about series expansions. This seems to be pretty key.

mducle: Ah sorry the $`\sqrt{2S}`$ terms refer to the scalar "spin length" or total spin quantum number $S$ rather than the vector $`\mathbf{S}=(S^x, S^y, S^z)`$. I'll make that clear in the text.

lucas-wilkins: You still need to square root the bs though,

lucas-wilkins (approved):

docs/developers/design/001_linear_spinwave_theory.md
mducle: @lucas-wilkins Ah, I get your meaning now - it's the $`S+ = b\sqrt{2S - b^{\dagger}}`$ term you mean.

Right, so in our particular case we only ever calculate the _square_ of these operators in the $`\mathbf{S}_i \mathrm{J}_{ij} \mathbf{S}_j`$ and $`\mathbf{S}_i \mathrm{A}_{i} \mathbf{S}_i`$ terms and in the Zeeman term we take the field to be along $z$ and use only the $S^z$ operator.

If we ever need to calculate the matrix elements of odd powers of these operators, then we would need to perform a Taylor expansion of the square root - which is tricky and requires partial differentiation of the operators. However, the authors [of this paper](https://doi.org/10.1103/PhysRevResearch.2.043243) worked out a way around this.

PR #13, opened 2024-11-05
Draft design decision for macroscopic structure 

### LINKED ISSUE #11: Add design decisions draft for experiment setup
Design decisions for how to handle
* Q scanning
* Twinning
* Powders

lucas-wilkins:
How are we going to handle macroscopic structure

### DIFF
--- a/docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
+++ b/docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
@@ -13,13 +13,11 @@ There is a question about whether it should be the location for the option to ap
 
 ## Decision
 
-Create a class representing the macroscopic structure that is responsible for working out the appropriate inputs for the
-main calculation, and constructing the measurable output.
+We will create a base `Sample` class with derived `TwinnedSample` and `PowderSample` classes which will have methods for preprocessing the input **Q**-vector list and post-processing the output spectra from `spinwave()` for the case of twinned crystals and powder (polycrystalline) samples. The base class will be used by default for untwinned crystals.
 
 ## Status
 
-Proposed
-
+Accepted
 
 ## Consequences
 
--- a/docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
+++ b/docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
@@ -1,0 +1,26 @@
+# A Structure for Describing Macroscopic Structure
+
+## Context
+
+The main job of spinW is to calculate the energy levels of a magnetic system over a range of q values. This
+task is relatively self-contained and well defined by just the material's cystallographic and magnetic properties.
+
+However, there are a number of things that are more properly understood as a property of a particular experiment. 
+The main features are to include here is whether or not the crystal is twinned or is a powder.
+
+There is a question about whether it should be the location for the option to apply neutron scattering selection rules.
+
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


mducle:
So the main task of SpinW is to compute the neutron intensity as a function of the momentum transfer **Q** and energy E - the main calculator actually takes as input the **Q** vectors and computes the magnon energies and cross-sections and then another part of the code then turns this into a computed intensity.

Now in the case of twinned crystals you would need the calculator to compute for several different sets of **Q** and then average the equivalent **Q** calculations from each set. For powders you need to generate a set of **Q** with the same magnitude $`|Q|`$ 
and average those. 

Whilst in both cases you do a preprocessing of the **Q** points list, feed this into the main calculator and then do some post-processing (averaging) the actual computations for the pre- and post-processing is quite different, so I don't know if it makes sense to have it in the same class? (Possibly a base class and two subclasses? E.g. an `ExperimentBase` and `TwinnedExperiment` and `PowderExperiment`?)

docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Decision
+
+Create a class representing the macroscopic structure that is responsible for working out the appropriate inputs for the
+main calculation, and constructing the measurable output.
+
+## Status
+
+Proposed
+
```
mducle:
```suggestion
We will create a base `Sample` class with derived `TwinnedSample` and `PowderSample` classes which will have methods for preprocessing the input **Q**-vector list and post-processing the output spectra from `spinwave()` for the case of twinned crystals and powder (polycrystalline) samples. The base class will be used by default for untwinned crystals.

## Status

Accepted
```

Edit: replaced `Crystal` with `Sample`.

mducle (changes requested):
During discussions on teams on Nov 6, we agreed to make it an abstract base class and two concrete classes, but didn't settle on a name - do you think `Crystal` is a good name? Alternatives are `Material`/`TwinnedMaterial`/`PowderMaterial` or `Experiment`/`TwinnedExperiment`/`PowderExperiment`.

docs/developers/adr/002_structure_for_describing_macroscopic_structure.md
RichardWaiteSTFC: Looks good, I'd be happy with this. I was wondering to what extent `TwinnedCrystal` can be represented by a collection/container of multiple `Crystal` base classes (perhaps it could still inherit from the `Crystal` base class - or maybe this breaks some rule of polymorphism - I don't know!). I think in general the S(Q,w) would need to be evaluated separately for each twin on the transformed Q-points (e.g. magnetic field may be along different inequivalent crystallographic axes in different twins). I guess the downside is that you might be more efficient to parallelise the spinwave loop over all the Q-points for all twins (as is done for powders) then assign then sum over twins after.

mducle: Essentially I was thinking of something in `spinwave()` along the lines of

```
q_list = crystal.parse_q(q_list)
Sab, omega = crystal.parse_sqw(calculate(q_list))
```

So in the base class `Crystal`, the `parse_q` and `parse_sqw` are no-ops (I assume that `calculate()` returns a tuple of `Sab, omega`).

lucas-wilkins: I'm thinking `Sample` is the right name

PR #16, opened 2024-11-06
Create 004_magnetic_lattice.md

### LINKED ISSUE #15: Design decision for MagneticLattice and MagneticPseudoLattice classes
Plan for how to represent the difference between commensurate and incommensurate systems.

lucas-wilkins:
Proposal for a separation between different kinds of magnetic structures

### DIFF
--- a/docs/developers/adr/004_magnetic_lattice.md
+++ b/docs/developers/adr/004_magnetic_lattice.md
@@ -14,7 +14,7 @@ In both cases the classes should implement a method to return the set of `zed` a
 
 In addition, it would be useful if the user can specify a set of propagation vector(s) and basis(es) in the explicit representation rather than every spin orientation in the unit cell, but this is an implementation detail to be discussed later.
 
-# Decision
+## Decision
 
 We will create two classes for handling the magnetic structure, a `CommensurateStructure` for the "explicit" representation and a `RotatingFrameStructure` for the "implicit" representation.
 
--- a/docs/developers/adr/004_magnetic_lattice.md
+++ b/docs/developers/adr/004_magnetic_lattice.md
@@ -8,13 +8,19 @@ The notion of commensurability maps on to this. Commensurate propagation vectors
 
 The plan is to make separate classes for these descriptions, provide some degree of conversion between the two, and associate appropriate calculations with each one.
 1) The explicit structure - every unique site is described explicity - explcit supercell, only (0,0,0), supports biquadratic
-2) The implicit structure - decribed by a unit cell, and a propagation vector - implicit supercell, supports incommensuate and true helical structures. The implcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
+2) The implicit structure - described by a unit cell, and a single propagation vector - implicit supercell, supports incommensurate and true helical structures. The impilcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
 
-Possible names could be ExplicitMagneticStructure/ImplicitMagneticStructure, MagneticLattice/MagneticPsudolattice, FullMagneticStructure/PartialMagneticStructure, etc.
+In both cases the classes should implement a method to return the set of `zed` and `eta` vectors needed for the spin wave calculations as detailed in [this doc](../design/001_linear_spinwave_theory.md#local-spin-directions).
+
+In addition, it would be useful if the user can specify a set of propagation vector(s) and basis(es) in the explicit representation rather than every spin orientation in the unit cell, but this is an implementation detail to be discussed later.
+
+# Decision
+
+We will create two classes for handling the magnetic structure, a `CommensurateStructure` for the "explicit" representation and a `RotatingFrameStructure` for the "implicit" representation.
 
 ## Status
 
-Proposed
+Accepted
 
 ## Advantages
 
--- a/docs/developers/adr/004_magnetic_lattice.md
+++ b/docs/developers/adr/004_magnetic_lattice.md
@@ -1,0 +1,22 @@
+# Classes for different kinds of magnetic structures
+
+## Context
+
+SpinW has different methods for dealing with different kinds of simulation, some simulations only work for systems decribed by a single (possibly complex) propagation vector, some (e.g. "biquadratic") only work
+when the propagation vector is (0,0,0). It seems that a useful distinction is between systems where every unique magnetic state is explicitly represented, and ones where it is implicitly defined. 
+The notion of commensurability maps on to this. Commensurate propagation vectors permit a finite (though potentially large) explicit representation.
+
+The plan is to make separate classes for these descriptions, provide some degree of conversion between the two, and associate appropriate calculations with each one.
+1) The explicit structure - every unique site is described explicity - explcit supercell, only (0,0,0), supports biquadratic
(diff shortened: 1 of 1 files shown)


docs/developers/adr/004_magnetic_lattice.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Context
+
+SpinW has different methods for dealing with different kinds of simulation, some simulations only work for systems decribed by a single (possibly complex) propagation vector, some (e.g. "biquadratic") only work
+when the propagation vector is (0,0,0). It seems that a useful distinction is between systems where every unique magnetic state is explicitly represented, and ones where it is implicitly defined. 
+The notion of commensurability maps on to this. Commensurate propagation vectors permit a finite (though potentially large) explicit representation.
+
+The plan is to make separate classes for these descriptions, provide some degree of conversion between the two, and associate appropriate calculations with each one.
+1) The explicit structure - every unique site is described explicity - explcit supercell, only (0,0,0), supports biquadratic
+2) The implicit structure - decribed by a unit cell, and a propagation vector - implicit supercell, supports incommensuate and true helical structures. The implcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
```
RichardWaiteSTFC: I like the separation of the two types of structure given the calculation proceeds differently for the two. I'm assuming (1) would also support k=0 with explicit supercell of `nExt = [1,1,1]` - I don't know if you would call this explicit supercell as it is the default.

I'm less sure about the names (but I can't come up with anything better!). By the SpinW definition we could call them (1) Commensurate and (2) Incommensurate. However as discussed this can cause confusion: a modulation of (1/2, 0, 0) with a single nuclear/parent unit cell would be classed as incommensurate in SpinW despite being a commensurate modulation by the common definition!

mducle: ~~In terms of names, what about `SuperCell` and `RotatingFrame`?~~
(Edit: sorry I just saw the issue with the name `MagneticLattice` and `MagneticPseudoLattice` - I presume `MagneticLattice` refers to the "explicit structure"? I think this is fine, but I don't like `MagneticPseudoLattice` - what about `RotatingFrameLattice` (I really like to have `RotatingFrame` somewhere in there!))

We could still have a `create_magnetic_structure` function for users similar to the current `genmagstr` method which would create a `MagneticLattice` or `RotatingFrameLattice` depending on its input, or we can separate out into different helper functions, like `create_helical_structure`, `create_fourier_structure`, `create_random_structure`, etc for the different [modes](https://github.com/SpinW/spinw/blob/73f604ab4d84084d872d1f5fdf46dbc54e14cdd7/swfiles/%40spinw/genmagstr.m#L51-L115) of `genmagstr`.

mducle: In terms of conversion between the two classes, you should always be able to convert from the implicit to the explicit representation (just construct a supercell with size determined by the inverse of the propagation vector).

However, conversion from an explicit to implicit representation will not always be possible. We could write a check for this (it would involve checking that the phase angle between spins in adjacent base unit cell is the same) but it might be easier to disallow this conversion as I think that probably only a minority of explicit cases can be converted to an implicit representation.

mducle:
```suggestion
2) The implicit structure - described by a unit cell, and a single propagation vector - implicit supercell, supports incommensurate and true helical structures. The impilcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
```

docs/developers/adr/004_magnetic_lattice.md
```diff
@@ -0,0 +1,22 @@
...
+
+Possible names could be ExplicitMagneticStructure/ImplicitMagneticStructure, MagneticLattice/MagneticPsudolattice, FullMagneticStructure/PartialMagneticStructure, etc.
```
mducle: After discussion on teams on Nov 6th 2024.

```suggestion
# Decision

We will create two classes for handling the magnetic structure, a `CommensurateStructure` for the "explicit" representation and a `RotatingFrameStructure` for the "implicit" representation.
```

docs/developers/adr/004_magnetic_lattice.md
```diff
@@ -0,0 +1,22 @@
...
+## Context
+
+SpinW has different methods for dealing with different kinds of simulation, some simulations only work for systems decribed by a single (possibly complex) propagation vector, some (e.g. "biquadratic") only work
+when the propagation vector is (0,0,0). It seems that a useful distinction is between systems where every unique magnetic state is explicitly represented, and ones where it is implicitly defined. 
+The notion of commensurability maps on to this. Commensurate propagation vectors permit a finite (though potentially large) explicit representation.
+
+The plan is to make separate classes for these descriptions, provide some degree of conversion between the two, and associate appropriate calculations with each one.
+1) The explicit structure - every unique site is described explicity - explcit supercell, only (0,0,0), supports biquadratic
+2) The implicit structure - decribed by a unit cell, and a propagation vector - implicit supercell, supports incommensuate and true helical structures. The implcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
+
```
mducle, line 12:
```suggestion

In both cases the classes should implement a method to return the set of `zed` and `eta` vectors needed for the spin wave calculations as detailed in [this doc](../design/001_linear_spinwave_theory.md#local-spin-directions).

In addition, it would be useful if the user can specify a set of propagation vector(s) and basis(es) in the explicit representation rather than every spin orientation in the unit cell, but this is an implementation detail to be discussed later.

```

docs/developers/adr/004_magnetic_lattice.md
```diff
@@ -0,0 +1,22 @@
...
+Possible names could be ExplicitMagneticStructure/ImplicitMagneticStructure, MagneticLattice/MagneticPsudolattice, FullMagneticStructure/PartialMagneticStructure, etc.
+
+## Status
+
+Proposed
```
mducle:
```suggestion
Accepted
```

mducle (changes requested):

docs/developers/adr/004_magnetic_lattice.md
```diff
@@ -0,0 +1,28 @@
...
+
+The plan is to make separate classes for these descriptions, provide some degree of conversion between the two, and associate appropriate calculations with each one.
+1) The explicit structure - every unique site is described explicity - explcit supercell, only (0,0,0), supports biquadratic
+2) The implicit structure - described by a unit cell, and a single propagation vector - implicit supercell, supports incommensurate and true helical structures. The impilcit structure requires the "trick" used in spinw to calculate intensities at +/- the propagation vector.
+
+In both cases the classes should implement a method to return the set of `zed` and `eta` vectors needed for the spin wave calculations as detailed in [this doc](../design/001_linear_spinwave_theory.md#local-spin-directions).
+
+In addition, it would be useful if the user can specify a set of propagation vector(s) and basis(es) in the explicit representation rather than every spin orientation in the unit cell, but this is an implementation detail to be discussed later.
+
+# Decision
```
mducle:
```suggestion
## Decision
```

mducle (approved):

PR #22, opened 2024-11-08
First steps to code for pyspinw

### LINKED ISSUE #21: Class structure outline
Create the basic structure for pyspinw

lucas-wilkins:
Initial work...

* Draft of the structure of pyspinw
* Decorator for checking array arguments
* Function to find a basis aligned with a given vector 

The find basis function returns a z aligned vector if the input is a vector of length 0, the other choice would be throwing an error, but I thought there might be edge cases where we want a basis for sites even if there is no moment.


pyspinw/dimensionality.py
```diff
@@ -0,0 +1,129 @@
+""" Tools for checking input shapes """
```
mducle, line 1: I know this file is trying to reproduce the SpinW parameters checker class but it still seems like a lot of code (>100 LoC!) for this. Also, I'm not sure that we really need such a class in Python as compared to Matlab because of Python's much richer handling of keyword arguments. Since we're targeting `v3.10` there are much greater support for type-hinting too - so for this particular case of dimensionality checking, can't we just use the [typing.Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) class to create aliases to some common sizes (like `3 x 3`, `N x 3`, etc) and then use regular type-hinting (with the [numpy.typing](https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing) module)?

pyspinw/hamiltonian.py
```diff
@@ -0,0 +1,30 @@
+""" Selection of different Hamiltonians """
+
+from pyspinw._base import Hamiltonian
+
+# pylint: disable=R0903
+
+
+class GeneralHamiltonian(Hamiltonian):
```
mducle, line 8: Could we change this name to `BilinearHamiltonian` please? (to make it clear compared to "biquadratic")

pyspinw/hamiltonian.py
```diff
@@ -0,0 +1,30 @@
...
+
+    """
+
+
+class HeisenbergMagnet(GeneralHamiltonian):
+    """
+    Heisenberg Ferromagnet Hamiltonian of the form H = - Σ J s.s
+
+    More explicitly, the equation is ... TODO
+    """
```
mducle, line 22: This class should be removed because it is handled by the above class (the `BilinearHamiltonian`).

pyspinw/hamiltonian.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class BiQuadratic(Hamiltonian):
```
mducle, line 25: Could we change this to `BiquadraticHamiltonian` please, to make its name of the same form as the above class?

pyspinw/structure.py
```diff
@@ -0,0 +1,24 @@
...
+
+class MagneticLattice(MagneticStructure):
+    """ A magnetic structure commensurate with the underlying crystal structure """
+
+
+    def add_propagation(self) -> "MagneticPsuedolattice":
+        """ Create a magnetic structure based on this with an additional propagation vector"""
+
+
+class MagneticPsuedolattice(MagneticStructure):
```
mducle, line 20: As mentioned in the standup, I don't like `Pseudo` here (typo by the way).

Could we change this `RotatingMagneticLattice` or `SingleQLattice` or something else, please?

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if rcond is None:
+        rcond = 3*np.finfo(vectors.dtype).eps # N * floating point epsilon
+
+    # Lengths, for checking for zeros and for normalising
+    lengths = np.sqrt(np.sum(vectors**2, axis=1))
+
+    #
+    # First basis vector will is just a normalised version of the input
+    #
```
mducle, line 22: Just a style preference but I don't think the extra empty comment lines are necessary here...

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Avoid zero divisions by using np.divide instead of /
+    zero_vectors = lengths < rcond
+    e_1 = np.divide(vectors, lengths.reshape(-1, 1), where=~zero_vectors.reshape(-1, 1))
+
+    # Assign zero vectors to z
+    e_1[zero_vectors, 0] = 0.0
+    e_1[zero_vectors, 1] = 0.0
+    e_1[zero_vectors, 2] = 1.0
```
mducle: I think that `e_1[zero_vectors,:] = np.array([[0., 0., 1.]])` will work here. Also in lines 43-45.

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
...
+
+    :param vectors: an n-by-3 numpy array containing the vectors to use to calculate the basis
+    :param rcond: tolerance on the calculation, default is 3 * machine precision
+    """
+
+    if rcond is None:
+        rcond = 3*np.finfo(vectors.dtype).eps # N * floating point epsilon
+
+    # Lengths, for checking for zeros and for normalising
+    lengths = np.sqrt(np.sum(vectors**2, axis=1))
```
mducle, line 18: What about using `np.linalg.norm(vectors, axis=1)` instead? It's shorter and more understandable...

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
...
+    e_1[zero_vectors, 2] = 1.0
+
+    #
+    # Second basis, cross with x-axis vector, unless its pointing that way already, then we choose y-axis explicitly
+    #
+
+    x_aligned = np.abs(e_1[:, 0] - 1) < rcond
+
+    x_vectors = np.zeros_like(vectors)
+    x_vectors[:, 0] = 1.0
```
mducle, line 38: I think this can be replaced by `x_vectors = np.tile([1., 0., 0.], (vectors.shape[0], 1))` which is a bit more obvious (to me at least!) and might be faster.

mducle (changes requested):
Looks good. Just a couple of optimisations in the basis code and some naming change requests.

~~I don't think the dimensionality checker class is necessary. I'd rather just stick to normal type-hinting.~~

pyspinw/structure.py
RichardWaiteSTFC: Yeh on the naming (sorry!) I also don't like the `Pseudo` - I also prefer `Structure` to `Lattice` just as a nod to previous spinw nomenclature.

pyspinw/dimensionality.py
RichardWaiteSTFC: I think I suggested checking vector/array sizes - I figured there might be a class of bugs where people converting their MATLAB scripts and not taking care of memory order etc..
It's nice to know this can now be done with type hinting, but I believe this doesn't do any runtime check, is that right? Maybe pydantic can do it?

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
...
+
+@dimensionality_check(vectors=(-1, 3))
+def find_aligned_basis(vectors: np.ndarray, rcond: float | None = None) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """ Find a set of orthonormal basis vectors aligned with the first being aligned to the input vectors
+
+    :param vectors: an n-by-3 numpy array containing the vectors to use to calculate the basis
+    :param rcond: tolerance on the calculation, default is 3 * machine precision
+    """
+
+    if rcond is None:
```
RichardWaiteSTFC, line 14: I thought `rcond` was something to do with whether it's OK to invert a matrix? This seems to be a tolerance for zero comparison? In any case I think doc-string could be clearer (I know thats nit-picking sorry!)

pyspinw/basis.py
```diff
@@ -0,0 +1,75 @@
+""" Functions for working with different basis vectors"""
+
+import numpy as np
+from pyspinw.dimensionality import dimensionality_check
+
+@dimensionality_check(vectors=(-1, 3))
+def find_aligned_basis(vectors: np.ndarray, rcond: float | None = None) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
```
RichardWaiteSTFC, line 7: This seems like a common thing to do (and I've done similar to what you do a few times myself) - so I had a quick look see if there's anything out there to do similar.
I've come up with this (but not checked thoroughly so no warranty!):
```
from scipy.spatial.transform import Rotation as rot
vec_aligned_to = [1,1,0]
rot, *_= rot.align_vectors(vec_aligned_to, [1,0,0])
rot.as_matrix()
```
which seems to do the trick?
```
array([[ 0.70710678, -0.70710678,  0.        ],
       [ 0.70710678,  0.70710678,  0.        ],
       [ 0.        ,  0.        ,  1.        ]])
```
it might be less efficient though and it doesn't work for zero length vectors so you'd still need to handle that.

pyspinw/dimensionality.py
mducle: Ah right... sorry yes, type hinting is only for static checking, it doesn't work at runtime.

pyspinw/hamiltonian.py
lucas-wilkins: Would have a different constructor

pyspinw/hamiltonian.py
lucas-wilkins: Probably, let's see what becomes of these classes.

pyspinw/hamiltonian.py
lucas-wilkins: I'm generally going with not having the type object directly in the name, probably the general one should just be `General` or `Bilinear`. The you'd access it though hamiltonian.General

pyspinw/structure.py
lucas-wilkins: Yeah, it will be done in the future.

pyspinw/basis.py
lucas-wilkins: done

pyspinw/basis.py
lucas-wilkins: done

pyspinw/basis.py
lucas-wilkins: You can invert a matrix if rcond is small. I'll call it tol instead.

pyspinw/basis.py
lucas-wilkins: It does what we want for a single (nonzero) vector, but when you give it an array it gives a single rotation that does as good as possible for all of them.

PR #31, opened 2025-03-21
Experiments in performance of matrix sqrt methods

### LINKED ISSUE #30: Benchmark LDL vs Cholesky
LDL is more general than Cholesky, but its probably at the cost of being slower, find out.

lucas-wilkins:
Results of some tests on replacing Choeskey with LDL, and an ADR.


mducle:
Sorry I think I forgot about this. It looks good though.

PR #37, opened 2025-05-06
32 gui prototype

### LINKED ISSUE #32: GUI Prototype
Prototype GUI for magnetic structure editing

lucas-wilkins:
GUI Prototype


pyspinw/gui/coordinateeditor.py
```diff
@@ -0,0 +1,267 @@
+import numpy as np
+from PySide6.QtCore import Signal
+from PySide6.QtWidgets import QWidget, QVBoxLayout, QGridLayout, QApplication, QLabel
```
mducle: Could we use the [qtpy](https://pypi.org/project/QtPy/) shim instead of PySide/PyQt directly please? The syntax is almost identical but would allow users to use PyQt6 or PySide6 (or even PyQt5/PySide2) as they prefer (or already have installed on their system).

pyspinw/symmetry/system.py
```diff
@@ -0,0 +1,218 @@
...
+            alpha=90,
+            beta=90,
+            gamma=90)
+
+    @property
+    def bravais_options(self) -> BravaisOptions:
+        return BravaisOptions(body_centered=True)
+
+
+class Rhombohedral(LatticeSystem):
```
mducle, line 209: I (and I assume most physicists) usually think in terms of the "Crystal Systems" rather than the "Lattice Systems" - they are identical except for the Trigonal/Hexagonal (Crystal Systems) and Rhombohedral/Hexagonal (Lattice Systems) division. The crystal systems are defined by the highest symmetry element whilst the lattice systems are defined by the lattice - thus trigonal crystals only have a three-fold rotation axis as the highest symmetry whilst rhombohedral lattices are defined by a single lattice parameter and lattice angle (which doesn't have to be any fixed value).

Now, in terms of saying something about what exchange interactions or anisotropy is allowed, the highest symmetry element is more useful than what lattice can be used for a particular structure. In addition, all trigonal, rhombohedral and hexagonal crystals can be described using the hexagonal lattice (a=b!=c, alpha=beta=90, gamma=120) but importantly, only some trigonal crystals can be described by a rhombohedral lattice but not all (and no hexagonal crystals [with 6-fold symmetry] can be described by a rhombohedral lattice.

All of this is a long-winded way to ask if you can change the definition to using "Crystal Systems" rather than "Lattice Systems" - and also to support both a hexagonal lattice for rhombohedral space groups like `R -3 m` in addition to a rhombohedral lattice.

mducle (changes requested):
It looks ok but it's a very large addition - in particular:

* I wonder if the `crystalviewer` module should be separate from PySpinW?
* Also, can't we use another existing library? Like [PyMol](https://github.com/schrodinger/pymol-open-source), [PyMolDyn](https://github.com/sciapp/pyMolDyn/) or [ase-GUI](https://wiki.fysik.dtu.dk/ase/ase/gui/gui.html#module-ase.gui)?

Also, there's really no tests of the GUI elements and although I'm not a huge fan of the model-view-presenter pattern, it does allow tests to be feasible. So maybe you can sort of refactor the view (pure GUI display) part out so that the call-backs and control logic (the "presenter") can be testable? Otherwise I'm just afraid that if someone else makes a change to the GUI later that things will break...

pyspinw/symmetry/system.py
lucas-wilkins: I've added an issue for this #45, as it will need to get added later, I have a plan for it

pyspinw/gui/coordinateeditor.py
lucas-wilkins: This sounds doable, will need to check a couple of things regarding Qt5 and Qt6 before a definite yes.

PR #39, opened 2025-05-07
Adds more lint rules (and lints accordingly)

alexhroom:
This PR adds some more lint rules that aren't captured in the `PL` ruleset, and lints the package accordingly.


lucas-wilkins (changes requested):
This is trying to merge a branch from the gui issue into main

lucas-wilkins (approved):

PR #43, opened 2025-05-08
Refactors spin-wave calculation for parallelism

alexhroom:
This PR refactors the spin-wave calculation so that it can be parallelised over q. Currently it seems that the parallelism slows things down, so I've put it behind a global flag `PARALLEL` while we're still testing things - I'll put some benchmarks together to see if it's faster for a large example. We also get a modest speed improvement as it avoids recalculating the `C` matrix for every q-value (it is q-independent) and also factors out the coupling-independent terms over sites.

### DIFF
--- a/pyspinw/calculations/spinwave.py
+++ b/pyspinw/calculations/spinwave.py
@@ -1,5 +1,6 @@
 """Spinwave Calculations"""
 
+import multiprocessing
 from concurrent.futures import wait, ProcessPoolExecutor
 from dataclasses import dataclass
 from enum import Enum
@@ -9,7 +10,7 @@ from scipy.linalg import ldl
 
 from pyspinw.checks import check_sizes
 
-PARALLEL = False
+PARALLEL = True
 
 
 # Disable linting for bad variable names, because they should match the docs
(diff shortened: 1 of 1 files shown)


pyspinw/calculations/spinwave.py
```diff
@@ -51,99 +54,99 @@ def spinwave_calculation(rotations: np.ndarray,
...
 
-        try:
-            sqrt_hamiltonian = np.linalg.cholesky(hamiltonian_matrix)
+        i = coupling.index1
+        j = coupling.index2
 
-            sqrt_hamiltonian_with_commutation = sqrt_hamiltonian.copy()
-            sqrt_hamiltonian_with_commutation[:, n_sites:] *= -1
+        A[i, j] += z[i, :] @ coupling.matrix @ z.conj()[j, :] * phase_factor
+        B[i, j] += z[i, :] @ coupling.matrix @ z[j, :] * phase_factor
```
mducle: I think that the expressions

```
z[i, :] @ coupling.matrix @ z.conj()[j, :]
```

and

```
z[i, :] @ coupling.matrix @ z[j, :]
```

are q-independent and that only `phase_factor` is q-dependent - so we should be able to factor the calculation of `A` and `B` further (e.g. remove the loop over `couplings`) to make it faster at the cost of looking a lot less like the equations.

alexhroom: we wouldn't be able to remove the loop over `couplings` because `phase_factor` is also dependent on the couplings (in particular the inter-site vector). I tried factoring it out (so the q-independent part of A and B are calculated when C is) and it didn't lead to a speedup, I think for this reason (in fact it may also lead to more array copies/writes)

mducle: So even if instead of using `z` we store two separate matrices:

```
Ac[i,j] = z[i, :] @ coupling.matrix @ z.conj()[j, :]
Bc[i,j] = z[i, :] @ coupling.matrix @ z[j, :]
```

it's not faster?

PS. We would have to be careful if a coupling is duplicated (two `coupling`s having the same `index1` and `index2` which is possible in SpinW as a convenience for users because in terms of the physics you think of a `coupling.matrix` which is proportional to the unit matrix `[[1, 0, 0], [0, 1, 0], [0, 0, 1]]` differently than one which is "antisymmetric" say, proportional to `[[0, 1, 0], [-1, 0, 0], [0, 0, 0]]` so that if you have a `coupling.matrix` of `[[1, 2, 0], [-2, 1, 0], [0, 0, 1]]`, in terms of the physics it makes more intuitive sense to decompose it into two `coupling`s)... so it might just be less buggy to leave the code as it is rather than try to eke out a bit of extra performance...

alexhroom: yeah, duplicate couplings cause issues - we only want to multiply the summand of `A[i,j]` for a *specific* coupling by the phase factor, but if we've added up A and B for all couplings in advance then it gets tricky to do that correctly

mducle:
@alexhroom 

I changed the code to parallelise over chunks of `q_vectors` instead of each q_vector individually and got a ~2x speedup:

Serial Python:
```
+------+---------------------+-----------------+--------------------+------------------+
| n_q  | ferromagnetic_chain | antiferro_chain | kagome_ferromagnet | kagome_supercell |
+------+---------------------+-----------------+--------------------+------------------+
| 100  |       0.121337      |     0.079652    |      0.406363      |     3.542459     |
| 500  |       0.603566      |     0.388887    |      2.017321      |    17.651729     |
| 1000 |       1.205572      |     0.806470    |      4.064086      |    35.582254     |
+------+---------------------+-----------------+--------------------+------------------+
```
Parallel Python:
```
+------+---------------------+-----------------+--------------------+------------------+
| n_q  | ferromagnetic_chain | antiferro_chain | kagome_ferromagnet | kagome_supercell |
+------+---------------------+-----------------+--------------------+------------------+
| 100  |       0.652569      |     0.509835    |      0.987606      |     2.949164     |
| 500  |       1.210489      |     0.683389    |      1.765415      |     8.909384     |
| 1000 |       1.501121      |     0.904601    |      2.356991      |    15.226496     |
+------+---------------------+-----------------+--------------------+------------------+
```
Not that impressive for a 16-core machine :shrug:

pyspinw/calculations/spinwave.py
```diff
@@ -48,25 +51,56 @@ def spinwave_calculation(rotations: np.ndarray,
...
     # Create the matrix sqrt(S_i S_j)/2
     root_mags = np.sqrt(0.5*magnitudes) # S_i / sqrt(2)
     spin_coefficients = root_mags.reshape(-1, 1) * root_mags.reshape(1, -1)
 
+    # calculate the C matrix for h(q), which is q-independent
+    C = np.zeros((n_sites, n_sites), dtype=complex)
+    for coupling in couplings:
+        i, j = (coupling.index1, coupling.index2)
+        C[j, j] += spin_coefficients[j,j] * eta[i, :].T @ coupling.matrix @ eta[j, :]
+    C *= 2
```
alexhroom, line 69: @mducle how does this line up with the theory? I see in Toth and Lake:

```math
C(k)_{j, j} = \delta_{ij} \sum_l S_l \eta_i^T J_{il}(0) \eta_l
```

how is the sum over $l$ captured here?

mducle: @alexhroom sorry for the late reply...

I actually don't know... Neither the Toth+Lake paper nor its predecessor by [Petit](https://www.neutron-sciences.org/articles/sfn/pdf/2011/01/sfn201112006.pdf) (just before section 3.3 on page 7 or 111) says what the sum over $$l$$ should run over... Looking at the matlab code, I see that this term is computed using [accumarray](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L900) with the [indices](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L635) simply being $$i==j$$ over the [list of couplings](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L593) - e.g. it's just the sum over the couplings, rather than an additional sum over the $$\eta$$ vectors... 🤷 

So, I'm not sure what this equation really means either... sorry...

henrikjacobsenfys: They mention in the paper (just after Eq 51) how they do the sum. The equation doesn't include a sum over the eta vectors - it just selects the appropriate ones for each coupling. So it seems to me that this matches the theory just fine.

> . To calculate these (A, B,C), we run a summation over the _l_ index of the list of couplings. Each _l_ value is associated with an _(i, j)_ index representing the indices of the interacting atoms in the unit cell.

mducle: Doh! Should have read it more carefully!

alexhroom: cool, i'm happy with that

lucas-wilkins (approved):

PR #48, opened 2025-05-28
Adds more examples and a benchmarking script

alexhroom:
This PR refactors the examples so they each consist of a function which creates the structure, and then the actual run is done in the `if __name__ == "__main__"` block. This makes it easier to just grab the structures of the examples so they can be run in a benchmarking script (which has been added as `benchmark.py` in the folder.

This also adds two additional examples:
- `antiferro_chain.py`: the structure from https://spinw.org/tutorials/02tutorial, to have a simple example where the rotations are not all trivial
- `kagome_supercell.py`: the structure from https://spinw.org/tutorials/08tutorial, a 3x3 supercell of the Kagome lattice to test an example for a large number of sites

### DIFF
--- a/examples/raw_calculations/antiferro_chain.py
+++ b/examples/raw_calculations/antiferro_chain.py
@@ -1,0 +1,41 @@
+"""Example of an antiferromagnetic chain.
+
+See https://spinw.org/tutorials/02tutorial
+"""
+
+import numpy as np
+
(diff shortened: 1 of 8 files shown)


alexhroom:
Reverting the Rust changes and moving them to a separate PR so we can call this "done" (although of course it doesn't work for the Rust!) and get it reviewed/merged

lucas-wilkins (approved):

PR #49, opened 2025-06-03
Update pyproject.toml to add build system

mducle:
* Add ADR to use ~~meson~~ maturin
* Add LICENSE (GPL-3)
* Add CITATION.cff

Otherwise a lot of boilerplate...


alexhroom:
of course this depends on whether we end up using rust but I've been using the [`maturin`](https://github.com/PyO3/maturin) build system for the rust mockup, which is first-rate built for mixed Rust/Python projects and is far more flexible to develop those mixed projects in (including built-in ability to compile mixed rust/python packages and publish them to pypi straight through the build system). it also supports `cffi` for C bindings

CITATION.cff
```diff
@@ -0,0 +1,36 @@
...
+    given-names: "Richard"
+    orcid: "https://orcid.org/0009-0006-4521-7950"
+  - family-names: "Jacobsen"
+    given-names: "Henrik"
+    orcid: "https://orcid.org/0000-0001-6950-3990"
+  - family-names: "Wilkins"
+    given-names: "Lucas"
+    orcid: "https://orcid.org/0000-0002-0149-3287"
+  - family-names: "Room"
+    given-names: "Alex"
```
alexhroom:
```suggestion
    given-names: "Alex H."
```

lucas-wilkins:
I guess if we're unsure about the build system, maybe we leave the status as "proposed" for now

mducle:
Merging as it's been so long!

PR #52, opened 2025-06-09
Adds Rust version of spinwave calculation 

### LINKED ISSUE #50: Rust implementation of spinwave.py
currently in https://github.com/alexhroom/pyspinw-rust-mockup

alexhroom:
Adds a Rust version of the spinwave calculation. Fixes #50

This uses the `maturin` build system. In short, to compile (with [rust installed](https://rustup.rs/)):
- type `maturin develop` to compile a dev build
- type `maturin develop --release` to compile a release build

Note some small changes to the low-level data:
- `rotations` is now a list of arrays rather than a 3D array of arrays
- currently you must ensure that the datatype of any rotation/coupling matrix is complex, and that the datatype of the coupling inter-site vector is a float (NOT an integer!). I think I'd ideally like to be handling this at the frontend as handling it within Rust is a bit of a pain

### DIFF
--- a/examples/raw_calculations/ferromagnetic_chain.py
+++ b/examples/raw_calculations/ferromagnetic_chain.py
@@ -1,5 +1,5 @@
 import numpy as np
-from pyspinw.calculations.spinwave import spinwave_calculation, Coupling
+from pyspinw.rust import spinwave_calculation, Coupling
 
 def heisenberg_ferromagnet():
 
@@ -11,10 +11,10 @@ def heisenberg_ferromagnet():
     q_vectors = np.array([0, 1, 0]).reshape(1, 3) * q_mags
 
     # Single site
-    rotations = np.eye(3).reshape(1, 3, 3)
+    rotations = [np.eye(3, dtype=complex, order='F')]
     magnitudes = np.array([1.5])  # spin 3/2
-    couplings = [Coupling(0, 0, np.eye(3), inter_site_vector=np.array([0, 1, 0])),
-                 Coupling(0, 0, np.eye(3), inter_site_vector=np.array([0, -1, 0])),
+    couplings = [Coupling(0, 0, np.eye(3, dtype=complex, order='F'), inter_site_vector=np.array([0., 1., 0.])),
+                 Coupling(0, 0, np.eye(3, dtype=complex, order='F'), inter_site_vector=np.array([0., -1., 0.])),
                  ]
 
     energies = spinwave_calculation(rotations,
@@ -32,6 +32,6 @@ if __name__ == "__main__":
 
     # Note: we get complex data types with real part zero
 
-    plt.plot(q_mags, result.raw_energies)
+    plt.plot(q_mags, result)
 
     plt.show()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,20 @@
-# TODO: Fill this out fully
+[build-system]
+requires = ["maturin>=1.3,<2.0"]
+build-backend = "maturin"
 
 [project]
-name = "PySpinW"
+name = "pySpinW"
+requires-python = ">=3.8"
+classifiers = [
+    "Programming Language :: Rust",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+]
 dynamic = ["version"]
+
+[tool.maturin]
+features = ["pyo3/extension-module"]
+module-name = "pyspinw.rust"
 
 [tool.setuptools.dynamic]
 version = {attr = "pyspinw.__version__"}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,202 @@
+use std::f64::consts::PI;
+use std::{iter::zip, ops::Sub};
+
+use nalgebra::{stack, Const, DMatrix, DMatrixView, DVector, Dyn, Matrix3, Vector3};
+use num_complex::Complex;
+use numpy::{PyArray1, PyReadonlyArray1, PyReadonlyArray2, PyReadwriteArray2, ToPyArray};
+use pyo3::prelude::*;
+use rayon::iter::IntoParallelRefIterator;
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -1,9 +1,21 @@
...
+build-backend = "maturin"
 
 [project]
-name = "PySpinW"
+name = "pySpinW"
+requires-python = ">=3.8"
+classifiers = [
+    "Programming Language :: Rust",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
```
mducle, line 11: Does PyPy support Rust modules? (I thought it's only CPython)

src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+    // where g is a diagonal matrix of length 2n, with the first n entries being 1, and the
+    // remaining entries being -1. We do this by just multiplying the >n_sites columns of shc.
+    let mut shc: DMatrix<C64> = sqrt_hamiltonian.clone();
+    let mut negative_half = shc.view_mut((0, n_sites), (2 * n_sites, n_sites));
+    negative_half *= Complex::from(-1.);
+
+    if let Some(v) = (shc.adjoint() * sqrt_hamiltonian).eigenvalues() {
+        v.data.into()
+    } else {
+        panic!()
```
mducle: Can we add an error message here, like `Could not compute eigenvalues of the Hamiltonian`, please?

src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+use pyo3::prelude::*;
+use rayon::iter::IntoParallelRefIterator;
+use rayon::prelude::{IntoParallelIterator, ParallelIterator};
+
+pub mod ldl;
+use crate::ldl::ldl;
+
+// some convenience types and statics for complex arithmetic
+type C64 = Complex<f64>;
+static J: C64 = Complex::new(0., 1.);
```
oerc0122, line 16: Is it not against rust convention to have non-types as caps?

src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+    let r: Vec<Matrix3<C64>> = rotations
+        .into_iter()
+        .map(|m| -> Matrix3<C64> {
+            let mv: DMatrixView<C64> = m.try_as_matrix().unwrap();
+            mv.fixed_resize::<3, 3>(Complex::from(0.))
+        })
+        .collect();
+    let qv = q_vectors.into_par_iter().map(Vector3::from_vec).collect();
+
+    let c = couplings.par_iter().map(pyo3::Py::get).collect();
```
oerc0122, line 70: Do these all need to be collected or is that just for simplicity?

pyproject.toml
alexhroom: seems so for what we want: https://pyo3.rs/v0.9.2/pypy.html but i'll double check

src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    let mut A = DMatrix::<C64>::zeros(n_sites, n_sites);
+    let mut B = DMatrix::<C64>::zeros(n_sites, n_sites);
+
+    for c in couplings {
+        let phase_factor = ((2. * J * PI) * q.dot(&c.inter_site_vector)).exp();
+        let (i, j) = (c.index1, c.index2);
+
+        A[(i, j)] += (z[i].transpose() * c.matrix * z[j].conjugate()).into_scalar() * phase_factor;
+        B[(i, j)] += (z[i].transpose() * c.matrix * z[j]).into_scalar() * phase_factor;
```
oerc0122, line 146: It feels like z'cz* should have a linear algebraic simplification.

examples/raw_calculations/ferromagnetic_chain.py
```diff
@@ -11,10 +11,10 @@ def heisenberg_ferromagnet():
     q_vectors = np.array([0, 1, 0]).reshape(1, 3) * q_mags
 
     # Single site
-    rotations = np.eye(3).reshape(1, 3, 3)
+    rotations = [np.eye(3, dtype=complex, order='F')]
```
oerc0122, line 14: Why `'F'` order? Is rust not `'C'` Ordered?

oerc0122 (commented):
Overall looks good. Just a couple of queries

pyproject.toml
alexhroom: for the record this `pyproject.toml` was autogenerated by `maturin` (except for the ruff bits which I added)

src/lib.rs
alexhroom: the rust docs tend to use all caps for statics: see e.g. https://doc.rust-lang.org/rust-by-example/scope/lifetime/static_lifetime.html or https://doc.rust-lang.org/reference/items/static-items.html

src/lib.rs
alexhroom: unless I'm missing something about giving iterators to functions, it's a lot simpler this way - one of the things we're concerned about is avoiding using too much weird and difficult Rust as that'll make it harder to maintain this in future

src/lib.rs
alexhroom: I agree, and I lost a chunk of an afternoon the other week trying to find one, but I don't think there is one (and this is the form given in papers that describe the algorithm)

examples/raw_calculations/ferromagnetic_chain.py
alexhroom: `nalgebra` uses 'F' order, so we can either define it here or try to fix it in Rust when we do `try_as_matrix`. Note that these matrices are internal, so the user won't have to specify this themselves, so I think this is the path of least resistance

oerc0122 (approved):

examples/raw_calculations/ferromagnetic_chain.py
oerc0122: That's fine, just wanted to check.

alexhroom: just thought I'd explain in case I'd missed something or you had a better idea!

PR #57, opened 2025-06-20
Fleshes out the outline of data structures

### LINKED ISSUE #51: Flesh out data structures
Image

alexhroom:
Fixes #51

Rearranges the data structures and adds some (placeholder) methods.

### DIFF
--- a/pyspinw/instrument.py
+++ b/pyspinw/instrument.py
@@ -1,17 +1,22 @@
 """Instrument specification"""
 
+from typing import Callable
 from dataclasses import dataclass
 
+from numpy.typing import NDArray
+
+@dataclass
 class Instrument:
     """Defines an instrument
 
     This class is responsible for instrument specific details, including resolution, energy and q binning, etc.
     """
 
-    resolution: float
+    energy: NDArray
+    resolution: Callable[[float], float] = lambda energy: np.max(self.energy)*0.02
 
     @classmethod
-    def from_ResINS(instrument_name: str) -> Instrument:
+    def from_ResINS(instrument_name: str) -> 'Instrument':
         """Instantiate an instrument from ResINS.
 
         Parameters
--- a/pyspinw/instrument.py
+++ b/pyspinw/instrument.py
@@ -1,6 +1,11 @@
 """Instrument specification"""
+
+from dataclasses import dataclass
+
 class Instrument:
     """Defines an instrument
 
     This class is responsible for instrument specific details, including resolution, energy and q binning, etc.
     """
+
+    resolution: float
--- a/pyspinw/instrument.py
+++ b/pyspinw/instrument.py
@@ -19,4 +19,4 @@ class Instrument:
         instrument_name: str
             The name of the instrument.
         """
-        raise NotImplementedError
+        raise NotImplementedError("ResINS compatibility has not yet been implemented.")
--- a/pyspinw/instrument.py
+++ b/pyspinw/instrument.py
@@ -9,3 +9,14 @@ class Instrument:
     """
 
     resolution: float
+
+    @classmethod
+    def from_ResINS(instrument_name: str) -> Instrument:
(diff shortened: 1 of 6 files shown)


mducle (changes requested):
Looks good but could you add another class diagram to the [003-structure.md](https://github.com/SpinW/pySpinW/blob/30480f3b991366f126ddef111b589694db84eda9/docs/developers/design/003-structure.md) file, like:

````
```mermaid
graph TD;
    A[Experiment]==>B{Fit};
    C[Data]==>B;
    C-->D(Experimental Q);
```
````

And add some description text to connect it to the `Experiment` class diagram, or modify the current `Experiment` class diagram itself to include the `Data` class, like:

````
```mermaid
   graph TD;
       A[Experiment]==>B[Sample]==>C(((Hamiltonian / Input Structure)))==>D[Sites];
       C==>E[Couplings]-.->D;
       A-->F(Magnetic Field);
       A==>G[Instrument]-->H(Resolution);
       B-->Ba[Sample Type]-->Bb(Q Points);
       A==>I{Fit};
       J[Data]==>I;
       J-->K(Experimental Q);
```
````

File: pyspinw/instrument.py [hunk 1]
```diff
@@ -1,6 +1,22 @@
...
+    @classmethod
+    def from_ResINS(instrument_name: str) -> Instrument:
+        """Instantiate an instrument from ResINS.
+
+        Parameters
+        ----------
+        instrument_name: str
+            The name of the instrument.
+        """
+        raise NotImplementedError
```
lucas-wilkins: Think this should be `NotImplementedError()` (call the constructor). Not to be confused with `NotImplemented` which is called without the constructor and used by __dunder__ methods

lucas-wilkins (approved):
looks fine other than the thing I mentioned

pyspinw/instrument.py
alexhroom: you don't actually need the parentheses unless you're giving an error message (e.g. `NotImplementedError("TODO")`). you can try both in the Python REPL and see they both work - I've usually seen it without the parentheses in this sort of case

alexhroom:
@mducle I believe it'll block the merge until you re-review with an approval - let me know if there's anything not up to scratch in my changes to the class diagrams.

pyspinw/instrument.py
```diff
@@ -1,6 +1,22 @@
 """Instrument specification"""
+
+from dataclasses import dataclass
+
 class Instrument:
```
mducle, line 9: Shouldn't this be decorated with a `@dataclass`?

pyspinw/instrument.py
```diff
@@ -1,6 +1,22 @@
...
     """Defines an instrument
 
     This class is responsible for instrument specific details, including resolution, energy and q binning, etc.
     """
+
+    resolution: float
```
mducle: In most cases we would want the resolution to be the energy width as a function of energy - so to be most flexible I would rather have it be a `Callable` with a default value of `lambda energy: constant_energy_width` where `constant_energy_width` is computed as 2% of the maximum measured energy (the best resolution we can get with our instruments is 2%... usually it's more like 4-5%)... so maybe change this code to:

```python
from typing import Callable
import numpy.typing as npt
...
@dataclass
class Instrument:
...
    energy: npt.NDArray
    resolution: Callable[float] = lambda energy: np.max(self.energy)*0.02
```

I think the resolution parameter (which should be the energy width as a function of the energy should be a callable, and default to `lambda energy: constant_energy_width` rather than a float (which would be `constant_energy_width`) as the float is too restrictive.

Edit: (I don't know if the code I wrote above will actually work!)

mducle (changes requested):
Looks good - just a couple of question regarding the `Instrument` dataclass...

mducle (approved):
Looks good! Thanks!

Issue #58, opened 2025-07-01
Fix rust code on more complicated examples


alexhroom:
The PR #48 has thrown up some problems with the Rust (and Python) code. Some of these are fixed in that PR but I think it's worth pulling them out and doing it in a separate PR at this point!

alexhroom:
it hangs on `antiferro_chain` and `kagome_supercell`... for the former I'm finding that there is a difference in the LDL outcome that makes the L matrix such that the matrix being diagonalised is all zeroes!

```
in both languages, D = diag([2, 2, 0, 0])

in Python, L is
 [1,  0,  0,  0]
 [0,  1,  0,  0]
 [0, -1,  1,  0]
 [-1, 0,  0,  1]

whereas in Rust it is
[1,  0,  0,  0]
[0,  1,  0,  0]
[0, -1,  0,  0]
[-1, 0,  0,  0]
```

alexhroom:
it seems we're actually running up against two nalgebra bugs:
dimforge/nalgebra#1291 (which I've submitted a fix for at dimforge/nalgebra#1532)
dimforge/nalgebra#611

mducle:
Thanks for looking into this. Is the second bug related to the deflation you talked about on Monday? (Although one of the linked issues mentioned that the infinite loop resolves if a small delta is _added_ to zero elements...?)

Anyway, as discussed on Monday, what do you think about moving to another library which wraps BLAS/LAPACK instead? [ndarray](https://docs.rs/ndarray-linalg/latest/ndarray_linalg/) looks ok...? There are two crates which just wraps the Fortran but their syntax are a bit ugly...

Edit: Another thing to bear in mind is that we might want an explicit _serial_ linear algebra library when we parallelise over q-points since this will reduce contention compared to the case where both the solver and q-point calculations are parallel. If we could have the option of running the linear algebra solver in either serial (for small matrices/unit cells) or parallel mode that would be even better!

alexhroom:
@mducle Yes that's the second bug. And yes the delta thing is weird - I've also found you can resolve the loop by adding something like 0.00001 to the diagonal of the original matrix, which we could do here but I'm generally anti-fudge factor.

[...]

PR #64, opened 2025-08-27
Allow example language selection and add unit test for comparing implementations

### LINKED ISSUE #63: Add unit tests to ensure Python and Rust implementations of spinwave are equal

alexhroom:
This PR makes the examples more flexible:
- adds a `rust` option to the example structures which selects which language to use for the couplings.
- adds an option where if "py" (or "python", etc.) is given as an argument to an example it'll run in Python instead of Rust
- adds a unit test that compares the Rust and Python implementations (currently freezes, see #58)

Fixes #63

### DIFF
--- a/examples/raw_calculations/antiferro_chain.py
+++ b/examples/raw_calculations/antiferro_chain.py
@@ -2,19 +2,30 @@
 
 See https://spinw.org/tutorials/02tutorial
 """
+import sys
+
 import numpy as np
 
 try:
-    from pyspinw.rust import spinwave_calculation, Coupling
+    from pyspinw.rust import spinwave_calculation as rs_spinwave, Coupling as RsCoupling
+    RUST_AVAILABLE = True
 except ModuleNotFoundError:
-    from pyspinw.calculations.spinwave import spinwave_calculation, Coupling
+    RUST_AVAILABLE = False
 
-def antiferro_chain(n_q = 100):
+from pyspinw.calculations.spinwave import spinwave_calculation as py_spinwave, Coupling as PyCoupling
+
+def antiferro_chain(n_q = 100, rust = False):
     """Antiferromagnetic chain.
 
     We use a 2x1x1 supercell to capture the magnetic rotation periodicity.
     """
-    rust_kw = {'dtype':complex, 'order':'F'}
+    if rust:
+        rust_kw = {'dtype':complex, 'order':'F'}
+        Coupling = RsCoupling
+    else:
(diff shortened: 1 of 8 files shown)


examples/raw_calculations/antiferro_chain.py
```diff
@@ -2,19 +2,30 @@
 
 See https://spinw.org/tutorials/02tutorial
 """
+import sys
+
 import numpy as np
 
 try:
-    from pyspinw.rust import spinwave_calculation, Coupling
```
lucas-wilkins, line 8: Do you think we could abstract out some of this boilerplate?

alexhroom: i think the tension I was having was between adding this flexibility but without making the actual example structures abstract to the point of being hard to use - but I'll have another toy around and see what I can do

alexhroom:
note from meeting: the MATLAB tests use 1e-6 abstol OR 1% reltol as their tolerances. I'm going to go with just 1e-6 abstol for now and consider the reltol option if we have any problems with it.

alexhroom:
now we have the Rust working (#69) we can see what the actual difference between the libraries is: it seems the biggest difference is in the Supercell example, where we find 6.3% of the elements are mismatched with a maximum absolute difference of 8.55e-7

mducle:
Cool! Those differences seems quite reasonable.

alexhroom:
@lucas-wilkins are you happy with the tolerances? you had some reservations when we discussed this yesterday

mducle:
> note from meeting: the MATLAB tests use 1e-6 abstol OR 1% reltol as their tolerances. I'm going to go with just 1e-6 abstol for now and consider the reltol option if we have any problems with it.

The [Matlab tests](https://github.com/SpinW/spinw/blob/master/%2Bsw_tests/%2Bsystem_tests/systemtest_spinwave.m#L8-L9) are a bit lazy and written as a [base class with a single verify method](https://github.com/SpinW/spinw/blob/master/%2Bsw_tests/%2Bsystem_tests/systemtest_spinwave.m#L134) inherited by all the tests which uses the same criteria for both the calculated energies and intensities. I think the relative tolerance is only needed for the intensities which can change by orders of magnitude - especially near zero energy there is a divergence which means the calculated intensity can blow up massively. For the energies, which is just what we have here I think it's better to stick with just the absolute tolerance.

lucas-wilkins (approved):

PR #66, opened 2025-09-04
65 move away from pydantic classes

### LINKED ISSUE #65: Move away from pydantic classes
Pydantic is not great for user facing classes, it removes a lot of control over how you interact with them, and produces obscure errors.

Do serialisation differently and remove pydantic.

lucas-wilkins:
I've removed the use of pydantic and implemented serialisation separately, this is now documented in design doc 004


lucas-wilkins:
Missed a whole section

docs/developers/design/004_serialisation_methods.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+However, if we do this with the data structures in pySpinW, shared nodes will be duplicated. Instead we need to do something different. 
+Whilst doing this we would like to keep to a tree-like serialisation much as possible, as we only deviate a small bit.
+
+Contexts
+========
+
+The strategy used to serialise pySpinW data is to use "contexts". 
+When we come to a node that can be a duplicate, instead of serilising diretly, we interact with a context object:
```
mducle, line 45: Typo: `serilising diretly` -> `serialising directly`.

docs/developers/design/004_serialisation_methods.md
```diff
@@ -0,0 +1,60 @@
...
+
+```mermaid
+   graph TD;
+      A-->|serialise|B-->|serialise|D;
+      A-->|serialise|C-->|serialise|D;
+      D-->|json|context-->|reference|B-->|json|A;
+      D-->|json|context-->|reference|C-->|json|A;
+```
+
+TODO: Better diagram
```
mducle, line 60: Yes, please! (I don't understand this diagram!)

Especially how `context` fits in - surely `context` should depend an all objects `A`, `B`, `C` and `D` and not just `D`?

Looking at [the code](https://github.com/SpinW/pySpinW/blob/086c08d73ff4ded37323baf03633f5610a28521d/pyspinw/serialisation.py#L175) each serialisable object has its own context, and the base `.serialise` method will use this context, but then looking at [specific code](https://github.com/SpinW/pySpinW/blob/086c08d73ff4ded37323baf03633f5610a28521d/pyspinw/coupling.py#L130-L139) I see that in `Coupling`, which depends on `LatticeSite` and `CellOffsetCoercible`, that `Coupling.serialise()` calls its children's _private_ `LatticeSite._serialise` and `CellOffsetCoercible._serialise` methods with *its own* context, which avoids the duplication.

----

For the diagram, I think you're trying to describe a situation where class `A` is composed of classes `B` and `C` and each of `B` and `C` in turn is composed of class `D` - is that correct?

```mermaid
classDiagram
      A *-- B
      A *-- C
      B *-- D
      C *-- D
```

So in this case we would have a single context (belonging to `A`), which is then passed to the `_serialise` methods of `B`, `C`, `D`.

```mermaid
classDiagram
class A {
   context c
}
      A *-- B
      A *-- C
      B *-- D
      C *-- D
      A --> B : _serialise(c)
      A --> C : _serialise(c)
      B --> D : _serialise(c)
      C --> D : _serialise(c)
```

??

pyspinw/_base.py
```diff
@@ -3,122 +3,43 @@
 This is an abstract outline, the actual implementations are in different files
 
 """
+from dataclasses import dataclass
```
mducle: Do we need this import? (it doesn't seem to be used anywhere else in the code...)

pyspinw/anisotropy.py
```diff
@@ -1,53 +1,81 @@
...
-        super().__init__(site)
+    serialisation_name = "anisotropy"
 
-        self._a = a
+    @check_sizes(anisotropy_matrix=(3,3))
+    def __init__(self, site: LatticeSite, anisotropy_matrix: np.ndarray):
+        self._site = site
+        self._anisotropy_matrix = anisotropy_matrix
 
     @property
```
mducle, line 21: Just so I understand this - a lot of these classes seem to be init-only / immutable. In this case of `Anisotropy` the `anisotropy_matrix` is read-only as there's no `@anisotropy_matrix.setter` method. So, e.g. if you define a Hamiltonian and then want to change the anisotropy you'd have to replace the anisotropy object. (PS actually the [hamiltonian.py code](https://github.com/SpinW/pySpinW/blob/086c08d73ff4ded37323baf03633f5610a28521d/pyspinw/hamiltonian.py#L21-L22) doesn't currently have the anisotropy). This is the same with `Coupling`.

I guess this is a design decision? If so, we should document it in an ADR - I see that you briefly [mention this in the ADR for fitting](https://github.com/SpinW/pySpinW/blob/086c08d73ff4ded37323baf03633f5610a28521d/docs/developers/adr/005_interface_for_fitting.md?plain=1#L13) but it should probably be expanded... I understand that making `Anisotropy`, `Coupling` and `MagneticLattice` immutable makes thing easier to test/statically check - my concern is that when we do fitting we have to change/update parameters (possibly a lot) and if we have to create new objects each iteration that this could substantially affect performance.

Now in the fitting ADR, it says to use a `private method` for updating parameter - does this mean just directly accessing the single dunder field e.g. `_anisotropy_matrix` directly? (And would this just undo the benefit of making the object immutable?) I guess this might be the best we can do since Python doesn't have any notion of `private` methods or `friend` class (in C++ nomenclature...)

pyspinw/coupling.py
```diff
@@ -69,17 +298,48 @@ class XYCoupling(Coupling):
...
+
+    def __init__(self, site_1: LatticeSite, site_2: LatticeSite,
+                 j: float,
+                 cell_offset: CellOffsetCoercible = None,
+                 name: str=""):
+
+        self._j = j
+        self._coupling_matrix = np.eye(3)
+        self._coupling_matrix[0,0] = j
+        self._coupling_matrix[1,1] = j
```
mducle: This could be made more succinct by changing to `self._coupling_matrix = np.diag([j, j, 0])` similarly to the code in `XXZCoupling`.

pyspinw/instrument.py
```diff
@@ -1,6 +1,22 @@
 """Instrument specification"""
+
+from dataclasses import dataclass
```
mducle: This `dataclass` import is not used.

pyspinw/serialisation.py
```diff
@@ -0,0 +1,250 @@
...
+                    raise SPWSerialisationError(f"Expected to find key {key} in first parameter, "
+                                                f"only found {found_keys}")
+
+            return fun(*args, **kwargs)
+
+        return wrapper
+    return decorator
+
+
+class SPWSerialisationContextGroup:
```
mducle, line 54: Why is there a `SPWSerialisationContextGroup` _and_ a `SPWSerialisationContext` class? It looks like `SPWSerialisationContext` is a bit superfluous unless you want to have multiple `SPWSerialisationContextGroup`s for each `SPWSerialisationContext`? (And why would you want this?)

[Similarly with `SPWDeserialisationContextGroup` and `SPWDeserialisationContext`

mducle (changes requested):
A minor typo in the docs but mostly I'm just a bit confused about the architecture...

We also need a test of the nested serialisation (a check that it actually avoids duplicating a member object) and the reference passing - looks like this is only used by `LatticeSite`, so ideally we'd need a `test_latticesite_serialisation.py` file...?

docs/developers/design/004_serialisation_methods.md
lucas-wilkins: The context doesn't belong to anything really, it's just passed around. It's created when `serialise` is called, and then passed through the tree, and used to catch cases where there are possibly convergent branches. In the diagram they're instances, not classes. Basically, it lets you dynamically build a table of references.

pyspinw/_base.py
lucas-wilkins: `_base` is going to go, its just not quite gone yet.

pyspinw/anisotropy.py
lucas-wilkins: Yeah, you need to rebuild things when you change values. It's not going to be the bottleneck though, at least, when the systems get to a size where updating parameters isn't basically instant, the diagonalisation will be taking much much longer.

pyspinw/serialisation.py
lucas-wilkins: It's fairly likely we'll need another group at some point in the near future, keeping it general for now.

pyspinw/anisotropy.py
mducle: So updating a parameter in a fit means creating new `Anisotropy` or `Coupling`, etc objects?

pyspinw/serialisation.py
mducle: I guess my point is what are the groups, and why are they needed?

Just looking at the code in more detail, it looks like `site` is specifically for a `LatticeSite` (or its derived `ImpliedLatticeSite` class) and gets populated by the `LatticeSite._serialise()` method which is also called by `Anisotropy` and its derived classes at the moment. But I don't like hard-coding in properties from a higher level class in a lower level class like this. Also because of the `expected_keys` checking, does it means that _every_ object which creates a context from a dict will need to have a `site` (and whatever other group we later put in)?

If instead of just having a number as a unique id, if you prepend it with the class or object name (e.g. `site_1`) then there should be no collision and you can just put it straight into the context without needing a group. You could also use a `uuid4()`.

docs/developers/design/004_serialisation_methods.md
mducle: Bump.

pyspinw/serialisation.py
lucas-wilkins: The reason not to just use numbers is that its fragile, each one does have a unique ID in memory, which is an int, but that isn't what we want to serialise, otherwise, when we start loading/saving there will be all sorts of confusion. The serialisation context keeps track of the ids used for serialisation (which are in fact ints) but these are not the ones used in memory.

I disagree about using a dict instead of a structured object. I think its always safer to use a structured object. This is especially true when it comes to serialisation, currently we have a structure of serialised data that is concrete, we know what the fields should be, and they're explicitly named. If we use arbitrary dict keys, we could end up dumping things into files in the wrong place without any warning that somethings gone wrong.

I personally don't think its "hard-coding in properties of a higher level class", its more like a parallel structure that just happens to live in the serialisation file.

I don't think UUIDs will be an improvement. An int is fine, and easier for humans to read.

pyspinw/anisotropy.py
lucas-wilkins: Yeah, we can revisit this choice if you like, but for now yes.

mducle: Ok. I've created an issue for it [pyspinw#70](https://github.com/SpinW/pySpinW/issues/70)

pyspinw/serialisation.py
mducle: Ok. But we should document this in `004_serialisation_methods.md` - that the context object is composed of groups with defined names (also the name could be misinterpreted as a group of context objects rather than a group of keys... maybe rename it to `SPWSerialisationContextItem`?)

PR #69, opened 2025-09-09
Fix Rust by changing to `faer`

alexhroom:
This PR changes our Rust code to use the `faer` linear algebra library instead of `nalgebra`. This now successfully works for all of our examples (now passes the unit test added in #64), and also has several benefits:

- just generally seems to be a lot faster, I think because we're not having to do anything weird
- remove all of the hacks and extra algorithms (e.g. LDL) that I had to add in to make up for nalgebra's weirdness because faer has robust implementations
- because the faer eigensolver is better we don't need to use lapack bindings like in #59 (so don't need to worry about shipping and compiling lapack)
- faer is designed for large matrices (nalgebra was originally designed for computer graphics & 3D vector calculations but became more general) so hopefully we get a better speedup on larger systems
- and also it's just generally a much cleaner & easier to understand interface, so the code is a lot easier to read!

I've also separated out the 'pure Rust' spinwave calculation into `spinwave.rs` and `lib.rs` now only contains the actual Python bindings, which I think makes the code easier to pick apart and understand.

benchmarks (in parallel):
```
Python:
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| n_q  | heisenberg_ferromagnet | antiferro_chain | kagome_ferromagnet | kagome_antiferromagnet | kagome_supercell |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| 100  |        1.100310        |     0.656288    |      1.259882      |        0.548270        |     9.091611     |
| 1000 |        1.389890        |     1.144066    |      1.966586      |        1.503017        |    60.054960     |
| 5000 |        2.370992        |     2.022394    |      4.756256      |        4.113057        |    312.738876    |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
Rust (nalgebra):
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| n_q  | heisenberg_ferromagnet | antiferro_chain | kagome_ferromagnet | kagome_antiferromagnet | kagome_supercell |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| 100  |        0.015430        |     0.008893    |      0.015024      |        0.022338        |     0.228773     |
| 1000 |        0.018673        |     0.037555    |      0.099260      |        0.200462        |     2.236610     |
| 5000 |        0.078808        |     0.171370    |      0.415527      |        1.084438        |     9.642299     |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
Rust (faer):
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| n_q  | heisenberg_ferromagnet | antiferro_chain | kagome_ferromagnet | kagome_antiferromagnet | kagome_supercell |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| 100  |        0.003824        |     0.003737    |      0.006327      |        0.007670        |     0.084598     |
| 1000 |        0.018036        |     0.017578    |      0.038243      |        0.043796        |     0.716327     |
| 5000 |        0.073027        |     0.078108    |      0.174299      |        0.203319        |     3.585987     |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
```
where the `nalgebra` benchmarks are using the code in #59.


mducle (approved):
Looks good. Just tested it and found it works well for all the examples. The `faer` syntax also seems a bit less verbose than `nalgebra` since the default `mat` type is already dynamic.

PR #71, opened 2025-09-10
Adds CI to test compiled code

### LINKED ISSUE #68: Test with Rust in CI
Currently Rust isn't installed for the unit tests. This means we can't run the test added in #64 in CI.

alexhroom:
**Note: this PR sits on top of #64, not main!**

This PR adds a workflow to test the Rust code in CI. Fixes #68 and also fixes a small bug I spotted where the Python calculation gives a divide by zero error if there are fewer than 4 processes:
```python
n_proc = int(np.floor(multiprocessing.cpu_count() / 4))
# some lines... then in _get_q_chunks
int(np.floor(q_vectors.shape[0] / n_proc))
```
to fix this, I've just replaced the definition of n_proc with
```python
n_proc = max(int(np.floor(multiprocessing.cpu_count() / 4)), 1)
```

### DIFF
--- a/.github/workflows/rust-test.yml
+++ b/.github/workflows/rust-test.yml
@@ -1,0 +1,29 @@
+name: Rust tests
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.12"]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Set up Rust toolchain
+      uses: dtolnay/rust-toolchain@stable
+    - name: Install package and dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements-test.txt
+        pip install -r requirements.txt
+        pip install . -v
+    - name: Run tests with pytest
+      run: | 
+        # `pytest -m rust` only runs tests decorated @pytest.mark.rust
+        python -m pytest -m rust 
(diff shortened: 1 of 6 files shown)


alexhroom:
seems like it's failing because the pip package successfully installs, but then when you run `import pyspinw` it imports the local folder `pyspinw` rather than the installed package with the Rust code. not sure how to change the priority list here

mducle:
I think it's because you're running it in the repo root folder which has a `pyspinw` folder. The [Python module import priority](https://docs.python.org/3/tutorial/modules.html#the-module-search-path) always puts the current working folder first so you have to:

1. Change to another folder before running the test.
2. Rename the PySpinW module folder, e.g. move it into `src/pyspinw`.

I think a lot of projects adopt the structure in option 2. for this reason...

.github/workflows/rust-test.yml
```diff
@@ -0,0 +1,42 @@
...
+      uses: dtolnay/rust-toolchain@stable
+    - name: Install package and dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements-test.txt
+        pip install -r requirements.txt
+        pip install . -v
+    - name: Run tests with pytest
+      run: |
+        # go back a directory so it uses the installed pyspinw
```
mducle: Another possibility (I don't know that it would work!) is to run:

```
        pip install -e . -v
```
in line 29 above to create an "editable" installation. I _think_ this might copy the compiled rust library into the source `pyspinw` folder which would then mean that the test will work from just the root folder... maybe test locally first? (I don't have time right now but will try it tomorrow...)

mducle (approved):
Code looks good and CI test passes. Only thing is the weird changing directory requirements. I think we should add an issue to restructure the code (e.g. put both rust and python source into `src/`) to avoid having to do this.

.github/workflows/rust-test.yml
alexhroom: aha, that does work locally!

PR #73, opened 2025-09-22
62 finish coupling groups

### LINKED ISSUE #62: Finish coupling groups
Finish off the coupling groups that we're merged as a TODO.

lucas-wilkins:
Coupling groups, supercells, and associated changes.

This is still not quite finished, but I'm calling it good for now, probably better to merge at this point as I want to work on a different area (implementing the gen* methods) for a bit.


debug/symmetry_check_demos.py
```diff
@@ -0,0 +1,96 @@
...
+def display_warnings(warnings):
+    for offset in warnings:
+        print(f"Cell {offset}: {len(warnings[offset])} warnings")
+
+    for offset in warnings:
+        for warning in warnings[offset]:
+            print(warning)
+
+
+def check_moment_consistency_p1_example():
```
mducle, line 17: Shouldn't these checks in this `debug` file be in `tests`?

docs/developers/adr/007_expected_behaviour_for_site_changes.md
```diff
@@ -0,0 +1,47 @@
...
+
+This object, or it's usage, should be such that it obeys the following transitivity relationship, defined schematically as 
+
+   couplings(symmetry(sites), P1) <=> couplings(P1(sites), symmetry)
+
+That is, the couplings of sites defined by a symmetry group (the sites, and those implied by symmetry) should be the
+same as between a list of sites defined in P1 symmetry.
+
+The implementation of this is rather trivial, but it suggests a design choice. That is that the `CouplingGroup` object
+acts a list of all equivalent sites according to the symmetry group - as opposed to, for example,
```
mducle, line 33: typo: `acts a list` -> `acts on a list`

docs/developers/adr/007_expected_behaviour_for_site_changes.md
```diff
@@ -0,0 +1,47 @@
...
+creating couplings then applying a symmetry. It also means that the `CouplingGroup` object needs to be designed in
+such a way that it is insensitive to symmetry.
```
mducle, line 35: I'm not sure I understand what you mean here... Wouldn't specifying one `Coupling` between all equivalent sites be the same as applying the symmetry operations to a single instance of that `Coupling`?

Let's give a concrete example of a 2D square lattice with P4 symmetry with a 4-fold rotation axes at the origin and mirror planes at x=0 and y=0. 

* Case 1: we have two atoms: Atom 1 at `(0,0)` and Atom 2 at `(1/2,1/2)` and specify a single nearest neighbour `Coupling` from Atom 1 to Atom 2 - in this case creating a coupling from all equivalent Atom 1s and Atom 2s would be the same as applying the 4-fold rotation symmetry to the single coupling `(0,0) -> (1/2,1/2)`.
* Case 2: we have Atom 1 at `(0,0)` and Atom 2 at `(1/2, 0)` - by the 4-fold symmetry around `(0,0)` that means we must also have another Atom 2 at `(0, 1/2)` - so specifying a `Coupling` between Atom 1 and Atom 2 would also give the 4-fold symmetry. More generally if we have an Atom 2 at `(x, y)` then the 4-fold symmetry would give an equivalent atom at `(y, x)`, and the couplings would also have the 4-fold symmetry...

----

Another thing which I'm not sure we discussed is the option to have a set of `Coupling`s linking sites with the same distance regardless of symmetry - this is a behaviour of the current Matlab SpinW when the "special" spacegroup of `0` is given. E.g. in the above example of the square lattice, let's say that due to the position of a non-magnetic atom the spacegroup was P2 rather than P4, so we would only have a 2-fold rotation at the origin and no mirror planes - but if the only magnetic atom is at the origin then we'd still like the possibility to add 4 equivalent `Coupling`s at once: `(0,0)->(1,0)`, `(0,0)->(0,1)`, `(0,0)->(-1,0)`, `(0,0)->(0,-1)`.

docs/developers/adr/008_higher_than_required_symmetry.md
```diff
@@ -0,0 +1,35 @@
...
+We will take the approach that one does not need to represent every component of the system to run a calculation. 
+This allows a user to represent only the minimal number of atoms for performing a calculation, and also has 
+the advantage of simplifying the system.
+
+However, this leads to questions about how we treat symmetry. For example, 
+a P1 lattice with only one site will have much higher symmetry than P1 in the sense that
+there will be more groups that can leave it invariant than P1.
+
+Once we make the decision that the symmetry group we specify is a "lower bound", 
+there a problem arrises - one that exists in crystallographic group symmetry
```
mducle, line 13: typo: `there a problem arrises` -> `then a problem arises`

docs/developers/adr/009_symmetry_checking_couplings.md
```diff
@@ -0,0 +1,21 @@
...
+$H = \sum_{k=1}^{n_\text{couplings}} s_{i(k)} C_k s_{j(k)} $
+
+and one where the couplings for each pair of sites is lumped together, and we sum over (ordered) pairs of sites
+
+$H = \sum_{i =1}^{n_\text{sites}}\sum_{j =1}^{i} s_i \left(\sum_{C\in\text{couplings(i,j)}} C\right) s_j 
+= \sum_{i=1}^{n_\text{sites}}\sum_{j=1}^{i} s_i M_{ij} s_j $
+
+The second of these is important as it shows that 
+
+The symmetry
```
mducle, line 20: This ADR is incomplete...

docs/developers/design/004_serialisation_methods.md
```diff
@@ -58,3 +58,38 @@ When we come to a node that can be a duplicate, instead of serilising diretly, w
...
+once, in the `SPWSerialisable` class.
+These are public methods whose responsibility is to the create the serialisation context,
+provide it to the private methods `_serialise` and `_deserialise` which are recursive,
+and span the tree.
+It is also responsible for making sure that the context is serialised/deserialised.
+
+`_serialise` and `_deserialise` are not the only serialisation methods that 
+appear in the codebase.
+When a serialised class contains another, it can specify the type of contained class,
+and serialisation of the contained class depends on whether this class unique,
```
mducle, line 75: typo: `this class unique` -> `this class is unique`

docs/developers/design/004_serialisation_methods.md
```diff
@@ -58,3 +58,38 @@ When we come to a node that can be a duplicate, instead of serilising diretly, w
...
+which is called by the `_serialise` method. The serialise method will specify which
+subclass has been serialised, and return a dict of the form `{"type": ..., "data":... [shared keys]}`
+Each subclass will have the same `serialisation_name` but will have
+a different key for the other method, e.g. `transform_name`.
+
+Validation
+==========
+
+There remains a question about the degree of validation required on the saved data.
+In general, we take the approach that `.spw` files will rarely be edited by hand,
```
mducle, line 89: Is a `.spw` file a `json`? Shouldn't we just use `.json`?

pyspinw/couplinggroup.py
```diff
@@ -30,7 +32,8 @@ def accept(self, vector: np.ndarray) -> bool:
 class InDirectionFilter(DirectionalityFilter):
     """ Selects vectors in a given direction """
 
-    def __init__(self, direction: np.ndarray, max_dev_angle_deg: float):
+    def __init__(self, direction: ArrayLike, max_dev_angle_deg: float = 0.01):
```
mducle, line 35: Is `direction` meant to be a 3-vector? Didn't you have some decorator to check the sizes of inputs? (Or was that removed?)

pyspinw/couplinggroup.py
```diff
@@ -55,55 +59,59 @@ def accept(self, vector):
...
                     site_2=coupling.site_2,
                     cell_offset=coupling.cell_offset,
-                    **parameter_dict))
+                    **self.coupling_parameters))
+
+        # Apply the filtering
+        if self.direction_filter is not None:
+            kept_couplings = [coupling
+                              for coupling in kept_couplings
+                              if self.direction_filter.accept(coupling.vector(unit_cell))]
```
mducle, line 114: I don't understand the role of this filtering - Is it meant to represent the symmetry? If so, shouldn't we cover the possibility of multiple filters (given that the `*DirectionFilter`s classes implemented only cover `Coupling`s in a specified direction or plane)?

pyspinw/mutablestructure.py
```diff
@@ -57,14 +90,59 @@ def _update_symmmetry_sites(self):
...
+        """ Check the symmetry of this system and the system definition are consistent """
+        check_results = []
+
+        # Check site symmetry
+        all_sites = self._all_sites()
+        for i, site1 in enumerate(all_sites):
+            for site2 in all_sites[:i]:
+                if np.all(np.abs(site1.ijk - site2.ijk) < tolerances.SAME_SITE_ABS_TOL):
+
+                    if np.all(np.abs(site1.m - site2.m) > tolerances.SAME_SITE_ABS_TOL):
```
mducle, line 114: I think the greater than `>` here should be less than `<` given the error message ("`share the same position and moment`").

The thing is - would this actually be a conflict or just a duplicate which should be removed (e.g. produced by applying a set of symmetry operations)?

pyspinw/site.py
```diff
@@ -1,6 +1,7 @@
 """ Representation of sites (e.g. magnetic atoms) within a magnetic system"""
 
 import numpy as np
+from scipy.stats import goodness_of_fit
```
mducle, line 4: This import is not used anywhere in this file...

pyspinw/subsystems.py
```diff
@@ -0,0 +1,73 @@
...
+    for coupling in couplings:
+
+        # Group should be the same for both sites
+        node_index = unique_id_to_index[coupling.site_1.parent_site._unique_id]
+
+        group_index = nodes[node_index].group
+        coupling_groups[group_index].append(coupling)
+
+    # groups as list of tuples
+    return [pair for pair in zip(site_groups, coupling_groups)]
```
mducle, line 73: You could also use `return list(zip(site_groups, coupling_groups))`

pyspinw/symmetry/magneticstructure.py
```diff
@@ -0,0 +1,52 @@
...
+
+class MagneticStructure(SPWSerialisable):
+    """Base class for representations of the Magnetic Structures"""
+
+    serialisation_name = "structure"
+    """ Object to hold symmetry information together """
+
+    def __init__(self,
+                 space_group: SpaceGroup,
+                 magnetic_group: MagneticSpaceGroup,
```
mducle, line 16: Could we have a `| None` here which defaults to `P 1`, please?

pyspinw/symmetry/operations.py
```diff
@@ -301,7 +301,10 @@ def __call__(self, points_and_momenta: ArrayLike) -> np.ndarray:
         momenta = np.array(momenta)
 
         new_points = (point_operation @ points.T + translation).T % 1
-        new_momenta = momenta * self.time_reversal
+
+        # TODO: Work out which one of these is right
+        # new_momenta = momenta * self.time_reversal
+        new_momenta = ((point_operation @ momenta.T) * self.time_reversal).T
```
mducle, line 307: Does `momenta` here refer to the magnetic _moment_ vector **M**? If so, could we rename it to `moments`, please (since `momenta` is the plural of `momentum`)?

Also, I think that the correct operation is

```
new_moments = (( np.linalg.det(point_operation) * point_operation @ moments.T ) * self.time_reversal).T
```

According to [Wills](http://fermat.chem.ucl.ac.uk/spaces/willsgroup/wp-content/uploads/sites/6/2018/06/2017-Willls-Historical-introduction-to-the-symmetries-of-magnetic-structures-part-1-shah-1.pdf) section V and Fig 6 (more good references [here](http://fermat.chem.ucl.ac.uk/spaces/willsgroup/magnetic-structures/)).

pyspinw/symmetry/supercell.py
```diff
@@ -0,0 +1,409 @@
...
+from pyspinw.checks import check_sizes
+from pyspinw.serialisation import SPWSerialisable, SPWSerialisationContext, \
+    serialise_fraction_or_builtin, deserialise_fraction_or_builtin, SPWDeserialisationContext, expects_keys, \
+    SPWSerialisationError, numpy_serialise, numpy_deserialise
+from pyspinw.site import LatticeSite
+from pyspinw.cell_offsets import CellOffset
+from pyspinw.util import rotation_matrix
+
+
+def _coerce_numeric_input(value: Fraction | float, max_denom=1000_000):
```
mducle, line 17: If we're going to be reading `cif` most files (e.g. [this example](https://rruff.geo.arizona.edu/AMS/CIF_text_files/00104_cif.txt)) only quotes coordinates to 4 or 5 decimal places, so we should reduce `max_denom` to `1000` or `10000`...

pyspinw/symmetry/supercell.py
```diff
@@ -0,0 +1,409 @@
...
+        if json["assured_commensurate"]:
+            return CommensuratePropagationVector(i, j, k)
+        else:
+            return PropagationVector(i, j, k)
+
+    def __eq__(self, other):
+        return self.i == other.i and self.j == other.j and self.k == other.k
+
+class CommensuratePropagationVector(PropagationVector):
+    """ Propagation vector with integer values"""
```
mducle, line 82: I know it's just a comment, but `integer` here should really be `rational`...

pyspinw/symmetry/supercell.py
```diff
@@ -0,0 +1,409 @@
...
+            supercell_type = json["type"]
+            further_data = json["data"]
+            return supercell_types[supercell_type]._deserialise_supercell(further_data, scale, context)
+
+        except KeyError as ke:
+            expected_names = ", ".join([f"'{key}'" for key in supercell_types])
+            raise SPWSerialisationError(f"Expected transform type to be one of {expected_names}") from ke
+
+
+class TrivialSupercell(Supercell):
```
mducle, line 266: I don't like this name... and also the comment says that it is "`a single unit cell`", but is this really the case? E.g. if you construct it as `TrivialSupercell(scaling=(2,2,2))` you'd actually get a 2 x 2 x 2 supercell not a single unit cell (which is 1 x 1 x 1)... Now, it's not clear in this code or the examples, but are the moments in the "next" cell just a copy of the moments in the base unit cell, or can they be different (and so must be specified by the user)? If the former then maybe it can be called a `ReplicatedSupercell` or `TiledSupercell` or `CopySupercell`? If the later, then it corresponds to the `genmagstr('mode', 'direct', ...` in Matlab so maybe it can be called `DirectSupercell`?

pyspinw/symmetry/supercell.py
```diff
@@ -0,0 +1,409 @@
...
+    @staticmethod
+    def _deserialise_supercell(json, scale, context):
+        transforms = [(PropagationVector._deserialise(part["vector"], context),
+                       SupercellTransformation._deserialise(part["transform"], context))
+                      for part in json]
+
+        return TransformationSupercell(transforms, scale)
+
+
+class SummationSupercell(CommensurateSupercell):
```
mducle, line 372: Maybe `FourierSupercell` is a better name, since the relation defining it is a Fourier series...

tests/serialisation/test_supercell_serialisation.py
```diff
@@ -0,0 +1,57 @@
...
+
+    assert isinstance(deserialised, TrivialSupercell)
+
+    assert deserialised._scaling == supercell._scaling
+
+def test_summation_supercell_serialisation():
+    """ Check that summation supercells serialise correctly """
+    vectors = [CommensuratePropagationVector(0, 0, 1/2),
+               CommensuratePropagationVector(1/3, 1/3, 1/3)]
+    supercell = SummationSupercell(vectors, scaling=(1,3,5))
```
mducle, line 22: It's not very clear to me how the `scaling` argument for these supercell works - I assume that to get an actual magnetic structure we would need to specify the basis (the magnetic moments) - so with `scaling=(1,3,5)` you would need a `15 x N` array where `N` is the number of magnetic atoms in the structural unit cells? Now, does the propagation vector then apply on top of this `1 x 3 x 5` supercell or do you apply the propagation vectors after populating the `1 x 3 x 5` supercell to create a "super-super"cell - I don't think that either cases are correct, though... I think only the case where `scaling=(1,1,1)` is correct from the point of the formulism...

mducle (changes requested):
Some minor typos and other changes are needed... Aside from these I've made some comments which are more meant to be discussion points than code changes...

pyspinw/symmetry/supercell.py
lucas-wilkins: yeah

pyspinw/symmetry/supercell.py
lucas-wilkins: Corresponds to "tiled" in the MATLAB I think

tests/serialisation/test_supercell_serialisation.py
lucas-wilkins: it creates a super-super cell, yeah its a bit pointless, was in the original matlab, might be helpful for testing.

docs/developers/adr/009_symmetry_checking_couplings.md
lucas-wilkins: WIP, continued in later PR

debug/symmetry_check_demos.py
lucas-wilkins: Its not an automated test. The debug checks are manual tests that can help debugging.

lucas-wilkins: The name `check_` is to avoid this confusion.

pyspinw/symmetry/supercell.py
lucas-wilkins: people do say/write integer to mean rational in this context though

pyspinw/symmetry/supercell.py
lucas-wilkins: I've no objection to this.

pyspinw/symmetry/supercell.py
lucas-wilkins: can do

docs/developers/adr/007_expected_behaviour_for_site_changes.md
lucas-wilkins: Updated the status to reflect its unfinished state

pyspinw/symmetry/supercell.py
lucas-wilkins: adding to issue about naming

pyspinw/symmetry/supercell.py
lucas-wilkins: added to rename list

lucas-wilkins:
All these things have been addressed in PR #85 and this has been merged into that branch

PR #74, opened 2025-09-22
Adds Zeeman (external field) term to calculation

### LINKED ISSUE #72: Add Zeeman term to spinwave calculation
Add the effect of external magnetic field (Zeeman term) to the spinwave calculation in both Rust and Python.

alexhroom:
This PR makes it possible to provide an external magnetic field to the calculation. This is done via a new dataclass called MagneticField which contains a `vector` and a `g_tensor` list of arrays. Fixes #72 

This PR also adds two examples:
- `antiferro_ef.py`, which is [tutorial 29](https://github.com/SpinW/spinw/blob/dc06a4bec2c44bcfde6baa630ecb6329fb4b68ba/tutorials/tutorial/tutorial29.m) for MATLAB spinW. This is the antiferro chain with added anisotropies and an external magnetic field applied.
- `ferromagnet_gtensor.py`, an example for a simple ferromagnetic chain with an explicit field and g-tensor specified.

### DIFF
--- a/pyspinw/constants.py
+++ b/pyspinw/constants.py
@@ -1,0 +1,4 @@
+"""Mathematical constants used by pySpinW."""
+
+# Bohr magneton in units meV/T
+MU_B = 0.05788382
--- a/src/spinwave.rs
+++ b/src/spinwave.rs
@@ -1,15 +1,11 @@
 use std::f64::consts::PI;
 
-use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Scale, Side};
+use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Side};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
 use crate::{Coupling, MagneticField, C64};
+use crate::constants::{J, SCALAR_J, MU_B};
 
-// for convenience
-// `Scale` is the faer scalar type used for matrix-scalar multiplication
-static J: C64 = C64::new(0., 1.);
-static SCALAR_J: Scale<C64> = Scale(J);
-static MU_B: f64 = 0.05788382;
 
 /// Run the main calculation step for a spinwave calculation.
 #[allow(non_snake_case)]
@@ -47,7 +43,7 @@ pub fn calc_spinwave(
             f.g_tensors
                 .into_iter()
                 .enumerate()
-                .map(|(i, t)| -0.5 * MU_B * f.vector.transpose() * t * etas[i].clone())
+                .map(|(i, t)| -0.5 * MU_B * f.vector.transpose() * t * etas[i])
                 .collect(),
         ),
         None => None,
--- a/src/spinwave.rs
+++ b/src/spinwave.rs
@@ -3,12 +3,13 @@ use std::f64::consts::PI;
 use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Scale, Side};
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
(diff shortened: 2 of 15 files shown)


pyspinw/constants.py
```diff
@@ -0,0 +1,4 @@
+"""Mathematical constants used by pySpinW."""
+
+# Bohr magneton in units meV/T
+MU_B = 0.05788382
```
mducle, line 4: We should probably use `scipy.constants` instead - the syntax is a bit unwieldy though:

```python
MU_B = scipy.constants.physical_constants['Bohr magneton in eV/T'][0] * 1000
```

Also, `scipy.constants` is a bit of a big table and importing it takes ~1s - so it might be better for us to have our own table... I don't know, I'm 50-50 :/

src/spinwave.rs
```diff
@@ -3,12 +3,13 @@ use std::f64::consts::PI;
...
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
-use crate::{Coupling, C64};
+use crate::{Coupling, MagneticField, C64};
 
 // for convenience
 // `Scale` is the faer scalar type used for matrix-scalar multiplication
 static J: C64 = C64::new(0., 1.);
 static SCALAR_J: Scale<C64> = Scale(J);
+static MU_B: f64 = 0.05788382;
```
mducle: I can see constants like this being needed in multiple calculations routines - would it be a big overhead to have a separate `constants.rs` module which is imported in, like `mod constants` and used as `constants::MU_B`?

mducle (approved):
Looks good. Some minor comments about the constants (bohr magneton).

mducle:
One of the strengths of SpinW was that it had a lot of examples for users to see and copy, but at the moment, the examples here are very opaque and geared towards testing the code. We should decide:

* Should test examples be easily understood and useable by users as templates for their own calculations?
* Or should we have two sets of examples - a "raw" set as currently defined geared towards testing and another set of examples made for users?

@RichardWaiteSTFC @lucas-wilkins what do you think?

What prompted this was seeing the `InternalClasses` wrapper - and the fact that we have to pass different classes into the examples function depending on whether we want python or rust calculations - I think it's ok for this PR, and adds flexibility when we need to add other helper classes. But, ideally, for users they'd define the same `Atom`s, `Coupling`s, `MagneticField`s and can then just set a `use_rust` or `use_python` flag, and an internal driver will pick the right classes (or would it be better to just use the Python classes in user facing applications and an internal converter to rust that is needed? @alexhroom what do you think?)

pyspinw/constants.py
alexhroom: I suggest we shelve that idea for now and if we need more constants in future we can switch to use `scipy.constants`?

src/spinwave.rs
alexhroom: I believe it would only add a (tiny) compile-time overhead. :)

alexhroom:
@mducle The impression I was working under is that we have the raw examples so that I can test the spinwave calculation routines while Lucas is still working on the API. Ideally, when the API is complete, these examples would be replaced with ones constructed using the user-facing classes and serve as both examples and test data, but right now they _have_ to be written using internal classes because we don't have user-facing ones!

As far as I'm aware the plan when pySpinW is complete is that we have the user-facing classes, and then we have an internal routine that converts them to the arguments used by the spinwave calculation. 

Maybe we should have an issue to put together an ADR on this?

lucas-wilkins (approved):

PR #78, opened 2025-09-23
Replaces `faer` `concat!` macro with block Hamiltonian function and makes minor refactors to the Rust code

### LINKED ISSUE #77: Remove use of `faer` `concat!` macro
The `faer` `concat!` macro isn't really designed for external use (per the lead maintainer) and a handmade function would be more efficient and stable.

alexhroom:
This PR fixes #77 by replacing the `faer` `concat!` macro, which is not recommended for external use, with a more stable own function implementation. This doesn't significantly change the function runtime:
```
old concat macro:
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| n_q  | heisenberg_ferromagnet | antiferro_chain | kagome_ferromagnet | kagome_antiferromagnet | kagome_supercell |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| 100  |        0.007045        |     0.006178    |      0.013309      |        0.016799        |     0.194486     |
| 1000 |        0.028488        |     0.029554    |      0.098945      |        0.126499        |     1.914994     |
| 5000 |        0.134682        |     0.141327    |      0.507639      |        0.659204        |    12.774269     |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
new block hamiltonian function:
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| n_q  | heisenberg_ferromagnet | antiferro_chain | kagome_ferromagnet | kagome_antiferromagnet | kagome_supercell |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
| 100  |        0.005785        |     0.005534    |      0.013091      |        0.014965        |     0.189698     |
| 1000 |        0.027250        |     0.030260    |      0.094214      |        0.121221        |     1.879567     |
| 5000 |        0.130085        |     0.144967    |      0.484230      |        0.613518        |    12.258631     |
+------+------------------------+-----------------+--------------------+------------------------+------------------+
```

this PR also makes some minor refactors to the Rust code:
- changes the internal functions to not have a preceding `_` (Rust denotes public functions with the `pub fn` syntax, so this isn't necessary and just makes the code harder to read for no reason)
- adds the #[inline] macro to internal functions (which can create some very minor performance improvements with basically no downside)
- avoids needing to re-iterate over vectors so much by removing an unnecessary iteration in the Python bindings over q-vectors, and combining the 3 calls to `get_rotation_component` into one `get_rotation_components` function
    - the `get_rotation_components` function also now avoids copying the 3rd component of the rotation matrices (eta) when it's only ever read
- changes some Vec's into slices in the single-Q spinwave function, which apparently helps the compiler optimise it...

### DIFF
--- a/src/spinwave.rs
+++ b/src/spinwave.rs
@@ -1,8 +1,7 @@
 use std::f64::consts::PI;
-use std::iter::zip;
 
-use faer::{unzip, zip, Col, Mat, MatRef, Scale, Side};
-use rayon::iter::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
+use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Scale, Side};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
 use crate::{Coupling, C64};
 
@@ -23,14 +22,8 @@ pub fn calc_spinwave(
 
     // decompose rotation matrices
     // in the notation of Petit (2011)
-    // eta[i] is the direction of the i'th moment in Cartesian coordinates
-    let z: Vec<Col<C64>> = zip(
-        get_rotation_component(&rotations, 0),
-        get_rotation_component(&rotations, 1),
-    )
-    .map(|(r1, r2)| r1 + (r2 * SCALAR_J))
-    .collect();
-    let etas = get_rotation_component(&rotations, 2);
+    // etas[i] is the direction of the i'th moment in Cartesian coordinates
+    let (z, etas) = get_rotation_components(rotations);
 
     // make spin coefficients array
     // so spin_coefficients[i, j] = sqrt(S_i S_j) / 2
@@ -43,7 +36,7 @@ pub fn calc_spinwave(
     let mut C = Mat::<C64>::zeros(n_sites, n_sites);
     for c in &couplings {
         C[(c.index2, c.index2)] += spin_coefficients[(c.index2, c.index2)]
-            * (etas[c.index1].transpose() * c.matrix.as_ref() * etas[c.index2].as_ref());
+            * (etas[c.index1].transpose() * c.matrix.as_ref() * etas[c.index2]);
     }
     C *= 2.;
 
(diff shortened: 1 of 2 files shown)


src/spinwave.rs
```diff
@@ -125,16 +117,47 @@ fn _spinwave_single_q(
...
+fn component_mul(a: &Mat<C64>, b: &MatRef<C64>) -> Mat<C64> {
     let mut product = Mat::<C64>::zeros(a.nrows(), a.ncols());
     zip!(&mut product, a, b).for_each(|unzip!(product, x, y)| *product = x * y);
     product
 }
+
+/// Combine the A, B, and C matrices into the Hamiltonian `h(q)` matrix.
+#[allow(non_snake_case)]
+#[inline(always)]
+fn make_block_hamiltonian(A: Mat<C64>, B: Mat<C64>, C: &Mat<C64>) -> Mat<C64> {
```
oerc0122, line 141: Does `C` need to be borrowed if it's never written?

alexhroom: `C` is already borrowed into the `spinwave_single_Q` function (where `make_block_hamiltonian` is called) - if `C` wasn't borrowed into `spinwave_single_Q` it would move into it, which is not allowed because closures (like the map over q_vectors) aren't allowed to capture variables. You would be correct if this function wasn't running inside of a map though!

src/spinwave.rs
```diff
@@ -76,15 +69,14 @@ fn _spinwave_single_q(
         B[(i, j)] += (z[i].transpose() * c.matrix.as_ref() * z[j].as_ref()) * phase_factor;
     }
 
-    A = _component_mul(&A, spin_coefficients);
-    B = _component_mul(&B, spin_coefficients);
+    A = component_mul(&A, spin_coefficients);
+    B = component_mul(&B, spin_coefficients);
 
     // create Hamiltonian as a block matrix (the stack! macro creates a block matrix)
```
oerc0122: `stack!` is no longer referenced here.

oerc0122 (commented):
This all looks sane and moving towards standard rust style. Just a minor thing with a comment out of date and I think this is probably good to go.

oerc0122 (approved):

PR #80, opened 2025-09-30
Make tests run only if lint passes

### LINKED ISSUE #79: Refactor CI to have other test depend on Lint passing
We currently have 3 workflows ([lint](https://github.com/SpinW/pySpinW/blob/6346fde7dcf9eb0375d7483031e58d58e238949b/.github/workflows/lint.yml), [test](https://github.com/SpinW/pySpinW/blob/6346fde7dcf9eb0375d7483031e58d58e238949b/.github/workflows/test.yml) and [rust-test](https://github.com/SpinW/pySpinW/blob/6346fde7dcf9eb0375d7483031e58d58e238949b/.github/workflows/rust-test.yml)) which are all run for each pull request.

We could enforce better practice and reduce CI load by making `test` and `rust-test` dependent on `lint` passing, so if the linting tests fails, the Python and Rust unit tests will not run.

[This stackoverflow answer](https://stackoverflow.com/a/64733705) shows how we can do this.

alexhroom:
Fixes #79. Merges the Rust and Python tests into one file for simplicity.


mducle (approved):
Looks good - thanks!

I didn't think that moving the Rust tests into the main test workflow was needed but I'm also not opposed to it so I think it's fine.

PR #83, opened 2025-10-08
Add spin-spin correlation function

### LINKED ISSUE #76: Add spin-spin correlation function to spinwave calculation
Add the ability to calculate spin-spin correlation in line with the MATLAB.

alexhroom:
Fixes #76 

This PR also tidies up the code to increase reusability by moving lots of reused functionality into a `utils.rs` file.


alexhroom:
conversation about outputs 8th dec: we want to do away with the `neutron` function and instead have `calc_spinwave` always return energies and sqw, optionally also returning sab if desired

docs/developers/design/001_linear_spinwave_theory.md
```diff
@@ -337,3 +326,60 @@ Instead SpinW uses one of two algorithms:
...
   Note that in this case SpinW can return (unphysical) imaginary magnon energies.
+
+## The spin-spin correlation function
+
+Once the Hamiltonian matrix has been diagonalised, the eigenvalues and eigenvectors can be used to compute the spin-spin correlation function
+which is directly related to the inelastic neutron scattering cross-section.
+
+Note that this calculation assumes that the eigenvalues (and their associated eigenvectors) are in decreasing order.
+
+For a matrix of eigenvectors $`U`$ and a $`2N \times 2N`$ diagonal matrix $`E = gL`$ where $`L`$ is the diagonal matrix of eigenvalues (the magnon energies),
```
mducle: I don't think actually define `g` in the above discussion of thr Hamiltonian - it would be good yo define it here.

docs/developers/design/001_linear_spinwave_theory.md
```diff
@@ -337,3 +326,60 @@ Instead SpinW uses one of two algorithms:
...
+```math
+\begin{align}
+\mathrm{Y}^{\alpha, \beta}_{ij}(\mathbf{q}) & = \sqrt{S_i S_j} \phi_{i,j}(\mathbf{q}) z_i^{\alpha} z_j^{*\beta} \\
+\mathrm{Z}^{\alpha, \beta}_{ij}(\mathbf{q}) & = \sqrt{S_i S_j} \phi_{i,j}(\mathbf{q}) z_i^{\alpha} z_j^{\beta} \\
+\mathrm{V}^{\alpha, \beta}_{ij}(\mathbf{q}) & = \sqrt{S_i S_j} \phi_{i,j}(\mathbf{q}) z_i^{*\alpha} z_j^{*\beta} \\
+\mathrm{W}^{\alpha, \beta}_{ij}(\mathbf{q}) & = \sqrt{S_i S_j} \phi_{i,j}(\mathbf{q}) z_i^{*\alpha} z_j^{\beta}
+\end{align}
+```
+
+Finally, we see that $`\delta(\omega - g_{ii}))`$ means the function is non-zero only at eigenvalues, so we can represent it as an array
```
mducle: I think an `\omega_i` is missing here after `g_{ii}`...

examples/raw_calculations/kagome_antiferro.py
```diff
@@ -45,6 +45,7 @@ def kagome_antiferromagnet(n_q = 100, classes = py_classes):
         rotation(-2 * np.pi / 3)
     ]
     magnitudes = np.array([1.0]*3)  # spin-1
+    positions = [np.array([0., 0., 0.]), np.array([1., 0., 0.,]), np.array([0., 1., 0.])]
```
mducle: Maybe put this on three lines like the kagome ferromagnet example above for consistency?

pyspinw/calculations/spinwave.py
```diff
@@ -181,3 +182,142 @@ def _calc_q_chunks(q_vectors: np.ndarray,
...
+    intensities = []
+    sabs = []
+
+    for q in q_vectors:
+        sqrt_hamiltonian = _calc_sqrt_hamiltonian(q, C, n_sites, z, spin_coefficients, couplings, Az)
+
+        sqrt_hamiltonian_with_commutation = sqrt_hamiltonian.copy()
+        sqrt_hamiltonian_with_commutation[n_sites:, :] *= -1  # This is C*K
+
+        to_diagonalise = np.conj(sqrt_hamiltonian).T @ sqrt_hamiltonian_with_commutation
```
mducle, line 254: Using `sqrt_hamiltonian.H` here might be marginally faster.

pyspinw/calculations/spinwave.py
```diff
@@ -181,3 +182,142 @@ def _calc_q_chunks(q_vectors: np.ndarray,
...
+
+        # note one can show:
+        # Y*[alpha, beta] = Y[beta, alpha]
+        # Z*[alpha, beta] = V[beta, alpha]
+        # V*[alpha, beta] = Z[beta, alpha]
+        # W*[alpha, beta] = W[beta, alpha]
+        # thus we only need to calculate one triangle and can fill in the rest by conjugation
+        for alpha in range(3):
+            for beta in range(alpha + 1):
+                z_alphas = z[:,alpha]
```
mducle: Might be able to save a (very small) bit by putting this line in the outer loop...

pyspinw/calculations/spinwave.py
```diff
@@ -181,3 +182,142 @@ def _calc_q_chunks(q_vectors: np.ndarray,
...
+                v_ab = np.conj(z_ab) * coefficients
+                w_ab = np.conj(y_ab) * coefficients
+
+                y_ab *= coefficients
+                z_ab *= coefficients
+
+                sab_blocks[alpha, beta] = np.block([[y_ab, z_ab], [v_ab, w_ab]])
+
+                if beta < alpha:
+                    y_ba = np.conj(y_ab.T)
```
mducle, line 297: As above might be able to save a bit of time with using `y_ba.H` etc. instead...

mducle (changes requested):
Looks ok... I've just got a few minor points, but I'm not sure why the tests are failing...

pyspinw/calculations/spinwave.py
alexhroom: doesn't exist - [`numpy.matrix.H` exists but _not_ `numpy.ndarray.H`!](https://numpy.org/doc/stable/reference/generated/numpy.matrix.H.html)

mducle: Doh!

alexhroom:
@mducle tests still fail but are a lot better after the changes to add a zero energy tolerance... and i've updated the examples to display the energies and sqw as two subplots on top of each other so it's a bit easier to see where the weird blow-ups are happening (it seems the only remaining massive one that won't fix itself is at small eigenvalues for the kagome example in Python)

mducle:
Thanks, I'll have a look at it later and will get back to you. Enjoy your break!

src/spinwave.rs
```diff
@@ -128,53 +245,222 @@ fn spinwave_single_q(
...
+            // V is conj(Z) and W is conj(Y)
+            let Yab = z_alphas.clone() * z_betas.adjoint();
+            let Zab = z_alphas.clone() * z_betas.transpose();
+            let Vab = Zab.conjugate().to_owned();
+            let Wab = Yab.conjugate().to_owned();
+
+            component_mul(&Yab, &coefficients);
+            component_mul(&Zab, &coefficients);
+            component_mul(&Vab, &coefficients);
+            component_mul(&Wab, &coefficients);
```
mducle: It looks like the results of these computations are not reassigned to `[YZVW]ab` - so it should be changed to `Yab = component_mul(&Yab, &coefficients)` or combined with the declarations above to `let Yab = component_mul(&(z_alphas.clone() * z_betas.adjoint()), &coefficients)`.

src/spinwave.rs
```diff
@@ -128,53 +245,222 @@ fn spinwave_single_q(
...
+    }
+
+    let eigendecomp = (sqrt_hamiltonian.adjoint() * shc)
+        .self_adjoint_eigen(Side::Lower)
+        .expect("Could not calculate eigendecomposition of the Hamiltonian.");
+
+    let eigvals: ColRef<C64> = eigendecomp.S().column_vector();
+
+    // we reverse the rows of U to match the nonincreasing order of eigenvalues in sqrt_E
+    let eigvecs: MatRef<C64> = eigendecomp.U().reverse_rows();
```
mducle: I think that if `reverse_rows()` is kept here that the `eigvals` should also be reversed - because the order of the `eigvecs` w.r.t. the `eigvals` are important.

In the equivalent Python code there is an `argsort` to get the indices of the `eigvals` which is then used to reorder both `eigvals` and `eigvecs` to the same order but I'm not sure how to do this in rust.

mducle (changes requested):
I tried to see how to fix the test discrepancies between Python and Rust calculations and found two parts where the rust calculations differ from the Python; however I still can't get the tests to pass - there is still some bug somewhere.

Looking at the values returned by Python and Rust codes I think that there is definitely some inconsistencies between the code and it's not just a matter of the intensities blowing up near regions where the eigenvalues goes to zero. 

Also, both Python and Rust calculated intensities is different from the Matlab - mostly the differences are small but at some Q points they are significant so I'm still hunting this issue down...

src/spinwave.rs
alexhroom: We do the argsort in the Python because the eigvals and eigvecs from Numpy aren't necessarily ordered _at all_, but in Rust they're always returned in nondecreasing order so we just have to reverse them.

src/spinwave.rs
alexhroom: whoops! good catch.

pyspinw/calculations/spinwave.py
```diff
@@ -26,158 +28,312 @@ class Coupling:
...
+    for q in q_vectors:
+        sqrt_hamiltonian = _calc_sqrt_hamiltonian(q, C, n_sites, z, spin_coefficients, couplings, Az)
 
         sqrt_hamiltonian_with_commutation = sqrt_hamiltonian.copy()
         sqrt_hamiltonian_with_commutation[n_sites:, :] *= -1  # This is C*K
 
         to_diagonalise = np.conj(sqrt_hamiltonian).T @ sqrt_hamiltonian_with_commutation
 
-        energies.append(np.linalg.eigvals(to_diagonalise))
+        eigvals, eigvecs = np.linalg.eig(to_diagonalise)
```
mducle: In this method we always assume that the part to be diagonalised is Hermitian - and to make the `antiferro_chain` test pass we need to use `eigh` here otherwise the eigenvectors are not always orthogonalised and will disagree with the Rust computation.

In addition, in order to pass the `antiferro_chain` test (see comment on Rust code below), we need to add a small delta to the diagonal, changing the line to:

```suggestion
        eigvals, eigvecs = np.linalg.eigh(to_diagonalise + np.diag(np.array(range(to_diagonalise.shape[0]))*1e-12))
```

src/spinwave.rs
```diff
@@ -128,53 +245,218 @@ fn spinwave_single_q(
...
+    }
+
+    let eigendecomp = (sqrt_hamiltonian.adjoint() * shc)
+        .self_adjoint_eigen(Side::Lower)
+        .expect("Could not calculate eigendecomposition of the Hamiltonian.");
+
+    let eigvals: ColRef<C64> = eigendecomp.S().column_vector().reverse_rows();
+
+    // we reverse the rows of U to match the nonincreasing order of eigenvalues in sqrt_E
+    let eigvecs: MatRef<C64> = eigendecomp.U().reverse_rows();
```
mducle: Right, sorry I think I told you the wrong thing... so what's important is that the _columns_ of `eigvecs` match up with the values of `eigvals` - so if we want to have it in the same order as `numpy`, we need a `reverse_cols()` in line 391 instead of `reverse_rows()` 

(Alternatively we could not do any reversing at all and revert the code in 398 back to `subrows_mut(0, n_sites)` - since we `sort` the values in the test anyway...)

src/spinwave.rs
```diff
@@ -111,7 +170,65 @@ fn spinwave_single_q(
             ldl.P().inverse() * l * sqrt_d.as_diagonal()
```
mducle: So I could only get the test to pass by using the `ldlt` method here instead of `lblt` - however, `ldlt` is not guaranteed to succeed so we'd have to fall back on `lblt` anyway... for the `kagome_ferromagnet` there seems to be an additional permutation which is needed but this only works for some Q-points. For others, the `L` computed by the decomposition is different from that compute by `numpy`'s `ldl` so it just gives a different `sqrt_hamiltonian` and then a different `Sab` (but the same eigenvalues, frustratingly!)... Anyway, the code I have which passes the test is:

```rust
        if let Ok(chol) = hamiltonian.clone().llt(Side::Lower) {
            chol.L().to_owned()
        } else {
            if let Ok(ldl) = hamiltonian.ldlt(Side::Lower) {
                let mut sqrt_d = Col::<C64>::zeros(hamiltonian.nrows());
                zip!(&mut sqrt_d, ldl.D().column_vector()).for_each(|unzip!(sqd, v)| *sqd = v.sqrt());
                ldl.L() * sqrt_d.as_diagonal()
            } else {
                let ldl = hamiltonian.lblt(Side::Lower);
                let l = ldl.L();
                let d = ldl.B_diag().column_vector(); // we're ignoring off-diagonals... this may be
                zip!(&mut sqrt_d, d).for_each(|unzip!(sqd, v)| *sqd = v.sqrt());
                // need to apply permutations: in Python scipy does this for you
                ldl.P().inverse() * l * sqrt_d.as_diagonal()
                let mut shm = Mat::<C64>::zeros(l.nrows(), l.ncols());
                perm::permute_cols(shm.as_mat_mut(), (ldl.P().inverse() * l * sqrt_d.as_diagonal()).as_mat_ref(), ldl.P().inverse());
                shm
            }
        }
     };
```

For the `kagome_ferromagnet` case `ldlt` always succeeds and the `lblt` section never gets executed. If we didn't have the `ldlt` section, it would fail for about 2/3 of the intensity values, despite the addition permutation of the columns of the `sqrt_hamiltonian` in the second last line (which needs additionally:

```rust
use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Par, Side, perm};
use faer::mat::{AsMatMut, AsMatRef};
```

at the start.

src/spinwave.rs
```diff
@@ -128,53 +245,218 @@ fn spinwave_single_q(
...
+                let Zba = Vab.adjoint().to_owned();
+                let Vba = Zab.adjoint().to_owned();
+                let Wba = Wab.adjoint().to_owned();
+
+                sab_blocks[(beta, alpha)] = block_matrix(&Yba, &Zba, &Vba, &Wba);
+            }
+        }
+    }
+
+    let eigendecomp = (sqrt_hamiltonian.adjoint() * shc)
```
mducle: The test failure on the `antiferro_chain` test is due to eigenvectors of degenerate energy levels not being unique or orthogonal; hence numerical differences between the Rust and Python code will give different `sab` values for the degenerate modes. We can add a small term to the diagonal of the matrix to be diagonalised to lift (numerically) the degeneracies to force the algorithm to make all eigenvectors orthogonal, by changing the above line to:

```suggestion
    // Adds a small delta to diagonal to ensure we don't have an exact degeneracies
    let diag_delta = Mat::<C64>::from_fn(shc.nrows(), shc.ncols(),
        |i,j| if i == j { C64::from((i as f64)*1e-12) } else { C64::ZERO } );
    let eigendecomp = ((sqrt_hamiltonian.adjoint() * shc) + diag_delta)
```

I found that a delta of `1e-12` was the smallest I could use and still get the test to pass, smaller values gave non-orthogonal eigenvectors and caused the test to fail - even then the test would fail unless I also added a relative tolerance of `1e-3`, or changed the absolute tolerance to `2e-3`.

tests/test_calc_impls.py
```diff
@@ -37,11 +37,13 @@
...
     """Compare Rust and Python spinwave calculation implementations."""
-    rs_results = rs_spinwave(*example(classes=rs_classes))
-    py_results = py_spinwave(*example(classes=py_classes))
+    rs_energies, rs_sqw = rs_spinwave(*example(classes=rs_classes))
+    py_energies, py_sqw = py_spinwave(*example(classes=py_classes))
 
     # we test to an absolute tolerance of 1e-6 in line with the MATLAB
-    np.testing.assert_allclose(np.sort(rs_results), np.sort(py_results), atol=1e-6, rtol=0)
+    np.testing.assert_allclose(np.sort(rs_energies), np.sort(py_energies), atol=1e-6, rtol=0)
+    np.testing.assert_allclose(np.sort(rs_sqw), np.sort(py_sqw), atol=1e-6, rtol=0)
```
mducle: Finally I found that I needed to change `rtol=1e-3` here to get the `antiferro_chain` test to pass, even with all the above changes...

mducle (changes requested):
Right, one last set of changes!

There two main issues causing the disagreement between Python and Rust (and hence test failures):

1. **Degenerate eigenvalues**: the eigenvectors for degenerate eigenvalues are not guaranteed to be unique or orthogonal by the eigendecomposition being computed in either Python or Rust - they can be any linear combination of each other and still satisfy the maths. Now, physically this is because we _cannot_ determine _separately_ the intensities of each degenerate mode - we can only measure their sum; And I think if we sum the intensities of degenerate modes we would find that the Rust and Python codes agree. (The [sw_omegasum](https://github.com/SpinW/spinw/blob/master/swfiles/sw_omegasum.m) function in Matlab SpinW does this). However, for our tests we are just comparing the raw values and this will be different depending on the numerics of `numpy` and `faer`.

2. There is something wrong with the `lblt` decomposition for the `kagome_ferromagnet` example... I can get slightly better agreement for some Q-points by doing an addition permutation of the points but I can only get the test to pass using the `ldlt` decomposition (which is not guaranteed to succeed)...

Anyway, below I've added suggestions for a quick fix for both these issues to get the test to pass... I'll let you decide if we keep these hacks and merge the PR or leave it open to find better solutions...

src/spinwave.rs
alexhroom: i don't think the order matters so much anyway because we don't return the eigenvectors to the user!

src/spinwave.rs
alexhroom: yeah I'd noticed the LDL discrepancies before... if I recall correctly the $L$ matrix produced by the LDL decomposition is not necessarily unique but I thought it'd come out in the wash (as it should still be mathematically kosher)

mducle (approved):
Looks good now, tests passed!

PR #93, opened 2025-11-28
Remove need for doubling up couplings

### LINKED ISSUE #90: Remove need for doubling-up couplings
Currently we need to manually specify couplings in both directions to get all the entries we need in the bosonic Hamiltonian. These can probably be automatically doubled up, but care needs to be taken to make sure that antisymmetric couplings (e.g. DM) are not cancelled.

Probably, we just need to add the transpose of each (i,j) coupling to the (j,i) coupling, but this needs to be checked (@mducle does this sound right?).

lucas-wilkins:
Removed the need to specify couplings in both directions along with some other improvements to the interface


mducle (approved):
Looks good and the examples are much cleaner!

PR #100, opened 2026-01-13
89 fix supercell hacks

lucas-wilkins:
Before Christmas, Supercells were disabled because there was some relatively complex code needed to set things up properly for them.

This implements that code.

PR #99, opened 2026-01-13
95 fill in sample classes

lucas-wilkins:
Sample classes added:

* Single Crystal
* Twin
* Multidomain
* Powder

Issue #96, opened 2026-01-06
Spinwave calculation interface updates


lucas-wilkins:
The following things could be updated for a cleaner interface between core code and user side.

* `MagneticField` is currently a class with the field, and g-factors for each site, refactor to put g-factor on sites, and convert class to an just array containing the `field` field.
* The output of calculations should really be wrapped up as an object
* Positions could be a an array, rather than a list of arrays

alexhroom:
The latter two of these were considered, but are mainly "I didn't want the headache":

> The output of calculations should really be wrapped up as an object

The annoying thing is that if you want to access the class from Rust, you need to write it _in_ Rust with the `#[pyclass]` attribute. But then if someone doesn't have access to the compiled Rust, they wouldn't have that class, so you'd have to write a backup second class in Python... 

if anything I'd say the bigger problem that needs solving there is "are we willing to assume that the user will always have the compiled Rust code" (they always _will_ unless they are using the package from the source - the PyPI package will always have the compiled Rust in it). If we assume yes, then there are several other simplifications we can make when it comes to the internal classes.

> Positions could be a an array, rather than a list of arrays

I mainly did this because a list of 1D arrays can be converted to Rust automatically as the `Vec<Vec<f64>>` native type, but a 2D array would need to have some custom code to convert the Python array to a Rust one. I also believe it's actually more efficient to store it as a list of arrays because you're only ever iterating over the rows (whereas in Python it's more efficient for it to be a 2D array so you can leverage Numpy!) and iterating over a `Vec` is more efficiently implemented than iterating over the rows of an array.

lucas-wilkins:
> so you'd have to write a backup second class in Python

Yeah, that was the plan though. The python side returns different kinds of tuples depending on the input, which is not great style (accept general, return specific) and this was the motivation for it.

> are we willing to assume that the user will always have the compiled Rust code

Good question, need to make a decision about this stuff soon (see #103)

mducle:
[Comment in PR107](https://github.com/SpinW/pySpinW/pull/107#discussion_r2696366191) on `experiment.py` changes to be implemented in this issue.

mducle:
> if anything I'd say the bigger problem that needs solving there is "are we willing to assume that the user will always have the compiled Rust code"

[...]

PR #107, opened 2026-01-15
Progressbar (with Supercell Hacks and Samples included)

lucas-wilkins:
Alex's new progress bar along with commits from #100 and #99


pyspinw/path.py
```diff
@@ -113,6 +113,29 @@ def format_plot(self, plt_or_fig=None):
...
         plt_or_fig.xticks(self.x_ticks(), self.x_tick_labels())
 
+class Path1D():
+    """ 1D Path, i.e. just values in absolute q """
+
+    def __init__(self,
+                 q_min: float = 0.0,
+                 q_max: float = 1.0,
+                 avoid_endpoints=True,
+                 resolution: int = 101):
```
mducle: Could we rename `resolution` to `npts` or `num` (like `linspace`), since it's actually the number of points. (Same with the `Path` class above).

pyspinw/anisotropy.py
```diff
@@ -37,45 +43,70 @@ def _deserialise(json: dict, context: SPWDeserialisationContext):
...
-class DiagonalAnisotropy(Anisotropy):
+class AxisMagnitudeAnisotropy(Anisotropy):
     """Anisotropy oriented with axes, but variable amount in x, y and z"""
 
-    @check_sizes(v=(3,), a=(1,), force_numpy=True)
-    def __init__(self, site: LatticeSite, a: float, v: np.ndarray = np.array([0, 0, 1])):
-        mag = np.sqrt(np.sum(v))
+    @check_sizes(direction=(3,), force_numpy=True)
+    def __init__(self, site: LatticeSite, a: float, direction: np.ndarray = np.array([0, 0, 1])):
+        mag = np.sqrt(np.sum(direction))
```
mducle: Shouldn't this be `np.sqrt(np.sum(direction**2))`? Or `np.linalg.norm(direction)`?

pyspinw/anisotropy.py
```diff
@@ -37,45 +43,70 @@ def _deserialise(json: dict, context: SPWDeserialisationContext):
...
 
-class DiagonalAnisotropy(Anisotropy):
+class AxisMagnitudeAnisotropy(Anisotropy):
     """Anisotropy oriented with axes, but variable amount in x, y and z"""
 
-    @check_sizes(v=(3,), a=(1,), force_numpy=True)
-    def __init__(self, site: LatticeSite, a: float, v: np.ndarray = np.array([0, 0, 1])):
-        mag = np.sqrt(np.sum(v))
+    @check_sizes(direction=(3,), force_numpy=True)
+    def __init__(self, site: LatticeSite, a: float, direction: np.ndarray = np.array([0, 0, 1])):
```
mducle, line 61: Could we change `a` to `A` (capitalised) since in equations it's usually a capital letter.

pyspinw/experiment.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, sample: Sample, instrument: Instrument | None = None):
     def calculate(self, input_q: list[np.ndarray], n_q: int, field: None = None) -> np.ndarray:
         """Calculate energies for the experiment."""
         generated_q = self.sample.generate_q(input_q, n_q, field, self.instrument.resolution)
-        energies = self.sample.hamiltonian.energies(generated_q)
+        energies = self.sample.hamiltonian.energies_and_intensities(generated_q)
 
         return energies
```
mducle, line 26: I know this is not used anywhere at the moment and it's not an error but `energies` is now a tuple of energies and intensities - maybe we should make this explicit in the code?

pyspinw/measurement.py
```diff
@@ -0,0 +1,21 @@
+""" Description of measurements """
```
mducle, line 1: I'm not sure the classes in this file (`measurement.py`) - and the related ones in `instrument.py` and `experiment.py` are needed right now (they are after all still just stubs), and also I'm not sure that these particular classes are what we want either.

Users with INS data will probably already use another program (Horace, Mantid, Mjolnir, Takin) to access and manipulate the data, and we don't want to redo what those programs do here - it would be better if we could interface to those programs instead of doing everything ourselves.

So for this PR, maybe we can remove the `measurements.py` file and revert the `instrument.py` changes?

pyspinw/sample.py
```diff
@@ -12,25 +20,227 @@ class Sample(ABC):
...
+
+    @check_sizes(field=(3,), allow_nones=True)
+    def energies(self,
+                 path: Path,
+                 field: ArrayLike | None = None,
+                 use_rust: bool = True) -> list[np.ndarray]:
+        """ Get the energies along a specified path """
+        return self._energies(path, field, use_rust)
+
+class Sample1D(Sample):
```
mducle, line 42: I'm not sure why this class is needed - couldn't we just have the `Powder` class inherit directly from `Sample`?

mducle (changes requested):
Looks good, I just have a few minor comments/requests. Also, I'd like to check what Euphonic does with its powder average first so will have some more comments on the averaging calculations `fibonacci`, `geodesic` and `spherical_integration` later...

pyspinw/measurement.py
lucas-wilkins: They are stubs, but we will need them to do calculations correctly.

pyspinw/sample.py
lucas-wilkins: Different methods available, general difference is that Sample1Ds have methods over 1D q-space paths, Sample3Ds have 3D have methods over 3D paths.

pyspinw/experiment.py
lucas-wilkins: yeah, this would come under #96

pyspinw/path.py
lucas-wilkins: Started changing this, then remembered there was a reason for it, namely that Path is based on some points, but `resolution` is not the number of these points, nor is it the number of returned points. I chose `resolution` to make this clearer, then in `Path1D` the convention follows `Path`.

pyspinw/anisotropy.py
lucas-wilkins: Done.

pyspinw/anisotropy.py
lucas-wilkins: I have made every (exposed) variable name in pySpinW conform to python conventions. It's also not what capital A is usually (a matrix), but is a scalar.

pyspinw/path.py
mducle: I think that `resolution` is close enough to number of points that it should be called `num` anyway - in fact it _is_ the _number of points per segment_ for `Path` and for `Path1D` it really is just the number of points. The only difference is whether you specify `avoid_endpoints` as true or not in which case it differs by `1`.

My main objection with using the word "resolution" is that it can be confused with the instrument resolution function. So I would really prefer this parameter to be called something different.

pyspinw/measurement.py
mducle: Yes, but my point is that these stubs are not needed in _this_ PR and it would be better to have them more fully implemented in another PR.

pyspinw/sample.py
mducle: Yes, but there's no code here... and how much of the code to be implemented would be used by `Powder`? I think it's better to remove this stub and implement this more fully in another PR later, and to have `Powder` inherit directly from `Sample` now.

pyspinw/calculations/fibonacci.py
```diff
@@ -0,0 +1,123 @@
...
+            return i
+
+    @staticmethod
+    def fibonacci_by_index(i):
+        """ Get the ith term of the Fibonacci series"""
+        # Two options, Binet's formula or evaluate series,
+        # Binets formula gives errors pretty quickly, so might
+        # as well use a series
+
+        return Fibonacci.fibonacci_series(i+1)[-1]
```
mducle: This implementation doesn't seem to be very performant - each call to `straddling_fibonaccis` needs two calls to `fibonacci_series` which recomputes the fibonacci series each time up to almost the same value. Also, we only really want `straddling_fibonaccis` and never really need the other static methods so why bother to have them? (Also also, why have a bunch of static methods in a class - why not just have them as independent functions?) Can't we just put everything into a single method and cache the series? E.g.

```python
from functools import cache

@cache
def fibonacci_series(n):
    return n if (n < 2) else fibonacci_series(n-1) + fibonacci_series(n-2)

def straddling_fibonaccis(n):
    lower_bound = int(np.log(n * _root_5)/np.log(_golden_ratio) - 1)
    lower_value = fibonacci_series(lower_bound + 1)
    last = (lower_bound + 1) if (lower_value < n) else lower_bound
    return fibonacci_series(last), fibonacci_series(last + 1)
```

mducle (changes requested):
Another change regarding the fibonacci.

The rest is fine - the Lambert projection is cool; I don't quite understand the geodesic code, will have to look at it more carefully but I think it's ok.

For reference the Euphonic discussion on this is summarised [on this page](https://github.com/pace-neutrons/pace-developers/blob/master/euphonic/design/05_powder_averaging_discussion.md) and the implementation is [here](https://github.com/pace-neutrons/Euphonic/blob/4c6b1d42efee08d1f530a33a30af3e3c9c125fca/euphonic/sampling.py)

pyspinw/measurement.py
lucas-wilkins: Yeah, I cut the plans for the Sample implementation short when I realised these classes need some more thought. They're just a sketch. I don't think we want unfinished things like this hanging around when it comes to a proper release, but there's a whole bunch of other cleaning up to do at that point as well. I find it easier to remember where I got to and what my plans were with stubs like this.

pyspinw/sample.py
lucas-wilkins: This is very much related to my previous comment.

lucas-wilkins: (on the other request)

pyspinw/path.py
lucas-wilkins: I'm not unsympathetic to this perspective.

I think `n_points` would have been fine if it was not for the other field `points` for which I cannot think of a better name for. I don't think we should go with `num` or `npts` as they are not in keeping with established style, which is `lower_case_underscore` and no abbreviations in exposed methods or for unexposed methods where there are no conflicts with existing conventions in other packages (e.g. `camelCase` for Qt, `plt` and `fig` for matplotlib stuff).

So, I would be more inclined to go with something like `[n_]points_per_segment` if the ambiguity is really unbearable.

pyspinw/calculations/fibonacci.py
lucas-wilkins: I try to avoid even memoised recursive definitions unless they're tail recursive and optimised for that. Stack attack.

Yeah, it could probably be cleaned up. I was playing around with other ways generating Fibonacci numbers and .

As for performance, its very quick without any optimisation. The Fibonacci series grows exponentially, so it's only ever going to be asking for less than around the 20th number. So we're adding about 50 numbers. 

The point generation is only called once per set up, so, the series generation will only be called twice. Probably quicker to just do some adding twice than once + cache.

I can optimise it, but I didn't really think it was worth it.

lucas-wilkins: Anyway, there's a much simpler way of doing it, which I've implemented now.

mducle (approved):

PR #109, opened 2026-01-19
108 fixes to power plotting

### LINKED ISSUE #108: Fixes to power plotting
Fix axes on powder plot

lucas-wilkins:
Fixed axes on powder plot, added a log intensity option

### DIFF
--- a/pyspinw/sample.py
+++ b/pyspinw/sample.py
@@ -149,6 +149,7 @@ class Twin(Multidomain):
 
 class ScalingMethod(Enum):
     """ Scaling methods for plots"""
+
     LINEAR = 'linear'
     LOG = 'log'
 
--- a/pyspinw/sample.py
+++ b/pyspinw/sample.py
@@ -1,6 +1,7 @@
 """Different Kinds of samples"""
 
 from abc import ABC, abstractmethod
+from enum import Enum
 
 import numpy as np
 from numpy._typing import ArrayLike
@@ -146,7 +147,10 @@ class Twin(Multidomain):
             domains=[CrystalDomain(np.eye(3), first_twin_fraction),
                      CrystalDomain(relative_rotation, second_twin_fraction)])
 
-
+class ScalingMethod(Enum):
+    """ Scaling methods for plots"""
+    LINEAR = 'linear'
+    LOG = 'log'
 
 class Powder(Sample1D):
     """Sample is a powder"""
@@ -216,6 +220,7 @@ class Powder(Sample1D):
                       max_energy: float | None = None,
                       n_energy_bins: int | None = None,
                       random_seed: int | None = None,
+                      scaling_method: ScalingMethod | str = 'linear',
                       show_plot: bool = True,
                       new_figure: bool = True,
                       use_rust: bool = True):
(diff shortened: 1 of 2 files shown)


pyspinw/sample.py
```diff
@@ -224,14 +230,26 @@ def show_spectrum(self,
                                    min_energy, max_energy, n_energy_bins,
                                    random_seed, use_rust)
 
+        if isinstance(scaling_method, str):
+            scaling_method = ScalingMethod(scaling_method)
+
+        match scaling_method:
+            case ScalingMethod.LOG:
+                zeros = data <= 0
+                min_value = np.log10(np.min(data[~zeros][:]))
```
mducle, line 239: Why not use `data>0` instead of `~(data <= 0)`? Also, maybe you need to use `nanmin` instead of `min`? I don't know that we can always guarantee that the calculations will not have NaNs...

pyspinw/sample.py
```diff
@@ -224,14 +230,26 @@ def show_spectrum(self,
...
+
+                data = np.log10(data)
+                data[zeros] =  min_value - 1
```
mducle, line 242: If you remove the `zeros` definition above (and use `min_value = np.log10(np.nanmin(data[data>0]))` you can change this to `data[np.isnan(data)] = ...`.

pyspinw/sample.py
```diff
@@ -224,14 +230,26 @@ def show_spectrum(self,
                                    min_energy, max_energy, n_energy_bins,
                                    random_seed, use_rust)
 
+        if isinstance(scaling_method, str):
+            scaling_method = ScalingMethod(scaling_method)
+
+        match scaling_method:
+            case ScalingMethod.LOG:
```
mducle, line 237: I can't think of any other types of scaling we'll need beyond linear and log, so not sure the whole `enum` and `match` is necessary - why not a simple `if`?

mducle (approved):
Looks good, just a couple of minor nitpicking comments...

pyspinw/sample.py
lucas-wilkins: Two reasons, enums are inspectable, so the type tells you what values are allowed (though for two, `bool` also tells you this) and writing a string or enum entry is clearer than `True`/`False`. Enums do the error handling too.

Beyond this, consider naming the log/linear option as a `bool`, you could do `log_scaling=True` or something similar, but what does `log_scaling=False` mean, specifically "not log scaling", but "not log scaling" is not logically the same as "linear" and you'd need to the context that these two options are the only ones to deduce that `log_scaling=False` means linear. So, really, the variable name should be `log_scaling_as_opposed_to_linear_scaling`, which is quite silly. Enum makes it clearer what your options are.

pyspinw/sample.py
lucas-wilkins: Because zeros makes more sense for a variable name, and I was using both that and its negation. I think we'd want data NaNs to show differently from zeros.

pyspinw/sample.py
lucas-wilkins: I avoid this kind of thing because it will produce a warning.

Issue #102, opened 2026-01-14
Ensure correct q scaling


lucas-wilkins:
The calculations currently assume that Q is in cartesian coordinates, but the inputs are designed as if they're in lattice units. Fix this.

Issue #98, opened 2026-01-08
Check Coupling Groups for duplicates


lucas-wilkins:
I'm not sure that interactions are not duplicated by coupling groups.

Does an X->X coupling make couplings with both (-1, 0, 0) and (1, 0, 0) offsets? it shouldn't.

lucas-wilkins:
There are two possible solutions to this, we need to decide which one is correct.

1) Constrain to an octant, e.g. only positive values in the offsets, I don't think this is right though
2) Constrain to a half-space, i.e. positive in only one direction, then some judicial choices around the x/y/z = 0 planes.

lucas-wilkins:
I think it needs to done like this:

<img width="859" height="457" alt="Image" src="https://github.com/user-attachments/assets/6efe7416-45e0-4ada-a569-de0a2cadc13f" />

mducle:
Right... actually I think the duplication is in treating `+0.0` differently from `-0.0`... so I think the issue is the zero line...

Taking an octant or a half plane in one or two axis definitely gives wonky results in the dispersion or structure factors `Sab`.

lucas-wilkins:
Just for clarity, the bit that expands the user facing `Coupling` objects to the input of spinwave should make two copies, that is correct. But the coupling generation code for the user facing objects definitely shouldn't make two couplings between the same site if the search distance is 1 LU, it should only make one

mducle:
Just reminder note to check this for higher spacegroup symmetry.

PR #110, opened 2026-01-23
Fix Sab / intensity calculations

mducle:
This PR ensures that output from PySpinW agrees with Matlab SpinW numerically. There were a couple of errors in the computations:

* A missing $`2\pi`$ in the intensity phase factor calculation
* The distance in the phase factor calculation of the Hamiltonian which we call the `inter_site_vector` is actually the vector between _unit cells_ (**G**) rather than the full vector between the sites.
* The transformation matrix T is actually `adjoint(L) \ U * sqrt(E)` rather than `L \ U * sqrt(E)` - in Matlab they use a Cholesky decomposition which yields an upper-triangular `K` whereas we use the LDL decomposition which yields a lower triangular `L` but the computation only works for an upper-triangular matrix.
* In `hamiltonian.py` the calculation of reverse couplings was missing the double counting factor of `1/2`
* The computation of distances for the couplings used the wrong transform (should not be transposed).
* The calculation of the orthogonal basis vectors `e1`, `e2`, `e3` had `e2` parallel to `[0,1,0]` if `e3` is parallel to `[1,0,0]` but the Matlab code had it parallel to `[0,0,1]` - and this caused the `kagome_supercell` calculation to fail (not sure why).
* Add code to remove duplicate couplings in the Hamiltonian calculations.

In addition, I changed the driver routines to correct the above mistakes; and also added more examples from the Matlab test suite.

However, there are two major discrepancies between the theory presented in Toth & Lake and the Matlab implementation:

1. There is an additional factor of `2` in the [Matlab Hamiltonian calculation](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L900) of the single-ion anisotropy terms (called `A20` / `D20` in the code, [computed in lines 618-9](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L618-L619) compared to the $`C^{i,j}`$ terms in eq (26) of the paper.
2. The scaling of the [Matlab Sab calculation (line 1096)](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L1096-L1097) is different to equation (47) of the paper's which gives a prefactor of $`1/(2N)`$ where $`N`$ is the number of atoms in the unit cell, but the Matlab code uses `1/(2*prod(nExt))` (factor of 1/2 is in [line 1066](https://github.com/SpinW/spinw/blob/master/swfiles/%40spinw/spinwave.m#L1066)) which is the number of unit cells in the supercell - e.g. the Matlab code computes the intensity _per unit cell_ whereas the theory gives it _per magnetic atom_.

In this PR I've added code to make the Python calculations consistent with the Matlab, but we might want to make it such that it agrees with the paper instead.

### DIFF
--- a/pyspinw/hamiltonian.py
+++ b/pyspinw/hamiltonian.py
@@ -28,6 +28,21 @@ from pyspinw.symmetry.supercell import TrivialSupercell
 
 logger = logging.Logger("pyspinw.hamiltonian")
 
+def omegasum(energy, intensity, tol=1e-5, zeroint=0):
+    """ Removes degenerate and ghost (zero-intensity) modes from spectrum """
+    energy, intensity = (np.array(energy), np.array(intensity))
+    en_out, int_out = (energy * np.nan, intensity * np.nan)
+    if tol > 0:
+        energy = np.round(energy / tol) * tol
+    if zeroint > 0:
+        energy[np.where(np.abs(np.real(intensity)) < zeroint)] = np.nan
+    for iQ in range(energy.shape[0]):
+        eu = np.unique(energy[iQ,:])
+        en_out[iQ,:len(eu)] = eu
+        for iE in range(len(eu)):
+            int_out[iQ, iE] = np.sum(intensity[iQ, np.where(np.abs(energy[iQ,:] - en_out[iQ, iE]) < tol)])
+    return en_out, int_out
+
 class Hamiltonian(SPWSerialisable):
     """Hamiltonian base class"""
 
@@ -252,32 +267,7 @@ class Hamiltonian(SPWSerialisable):
 
         return result[0], result[1]
 
-    def sorted_positive_energies(self,
-                                 path: Path,
-                                 field: ArrayLike | None = None,
-                                 use_rust: bool = True,
-                                 return_intensities: bool = False) -> list[np.ndarray]:
-        """ Return energies as series corresponding to q, sorted by energy """
-        energy, intensities = self.energies_and_intensities(path.q_points(), field=field, use_rust=use_rust)
-
-        energy = np.array(energy)
-
-        # Sort the energies
-        energy_index = np.argsort(energy.real, axis=1)
-
-        # return the top half (positive)
-        n_energies = energy.shape[1]
-
-        energy = np.take_along_axis(energy, energy_index, axis=1)
-        energies = [energy[:, n_energies - i - 1] for i in range(n_energies//2)]
-
-        if return_intensities:
-            intensities = np.take_along_axis(np.array(intensities), energy_index, axis=1)
-            intensities = [intensities[:, n_energies - i - 1] for i in range(n_energies//2)]
-            return zip(energies, intensities)
-        return energies
-
-    def plot(self,
+    def spaghetti_plot(self,
              path: Path,
              field: ArrayLike | None = None,
              show: bool=True,
@@ -294,8 +284,9 @@ class Hamiltonian(SPWSerialisable):
                 axs.append(fig.add_subplot(2,1,ii+1))
 
         x_values = path.x_values()
-
-        for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust, return_intensities=True):
+        energy, intensity = omegasum(*self.energies_and_intensities(path.q_points(), field=field, use_rust=use_rust))
+        n_mode = energy.shape[1]
+        for series in zip(*([v[:, n_mode - i - 1] for i in range(n_mode)] for v in (energy, intensity))):
             axs[0].plot(x_values, series[0], 'k')
             axs[1].plot(x_values, series[1])
         if 'log' in scale:
@@ -306,6 +297,26 @@ class Hamiltonian(SPWSerialisable):
 
         if show:
             plt.show()
+        else:
+            return fig
+
+    def sorted_positive_energies(self,
+                                 path: Path,
+                                 field: ArrayLike | None = None,
+                                 use_rust: bool = True) -> list[np.ndarray]:
+        """ Return energies as series corresponding to q, sorted by energy """
+        energy, intensities = self.energies_and_intensities(path.q_points(), field=field, use_rust=use_rust)
+
+        energy = np.array(energy)
+
+        # Sort the energies
+        energy = np.sort(energy.real, axis=1)
+
+        # return the top half (positive)
+        n_energies = energy.shape[1]
+        energies = [energy[:, n_energies - i - 1] for i in range(n_energies//2)]
+
+        return energies
 
     def energy_plot(self,
                     path: Path,
--- a/pyspinw/hamiltonian.py
+++ b/pyspinw/hamiltonian.py
@@ -181,7 +181,7 @@ class Hamiltonian(SPWSerialisable):
         lu2xyz /= np.sqrt(np.sum(lu2xyz**2, axis=1)).reshape(-1, 1)
         for index, site in enumerate(expanded._structure.sites):
             # TODO: Sort out moments for supercells
-            moments.append(site.xyz_moment(self.structure.unit_cell._xyz))
+            moments.append(site.xyz_moment(lu2xyz))
 
             positions.append(site.ijk)
 
--- a/pyspinw/hamiltonian.py
+++ b/pyspinw/hamiltonian.py
@@ -275,35 +275,50 @@ class Hamiltonian(SPWSerialisable):
             return zip(energies, intensities)
         return energies
 
+    def plot(self,
+             path: Path,
+             field: ArrayLike | None = None,
+             show: bool=True,
+             new_figure: bool=True,
+             use_rust: bool=True,
+             scale: str='linear'):
+        """ Create a spaghetti diagram with energy top and intensity bottom """
+        if new_figure:
+            fg, axs = plt.subplots(2, 1)
+        else:
+            fg = plt.gcf()
+            axs = fg.get_axes()
+            for ii in range(len(axs), 2):
+                axs.append(fg.add_subplot(2,1,ii+1))
+
+        x_values = path.x_values()
+
+        for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust, return_intensities=True):
+            axs[0].plot(x_values, series[0], 'k')
+            axs[1].plot(x_values, series[1])
+        if 'log' in scale:
+            axs[1].set_yscale('log')
+
+        path.format_plot(axs[0])
+        path.format_plot(axs[1])
+
+        if show:
+            plt.show()
+
     def energy_plot(self,
                     path: Path,
                     field: ArrayLike | None = None,
                     show: bool=True,
                     new_figure: bool=True,
-                    use_rust: bool=True,
-                    show_intensities=False):
+                    use_rust: bool=True):
         """ Create a spaghetti diagram """
-        if new_figure and show_intensities:
-            fg, axs = plt.subplots(2, 1)
-        elif show_intensities:
-            fg = plt.gcf()
-            axs = fg.get_axes()
-            for ii in range(len(axs), 2):
-                axs.append(fg.add_subplot(2,1,ii+1))
-        elif new_figure:
+        if new_figure:
             plt.figure("Energy")
 
         x_values = path.x_values()
 
-        if show_intensities:
-            for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust, return_intensities=True):
-                axs[0].plot(x_values, series[0], 'k')
-                axs[1].plot(x_values, series[1])
-            if axs[1].get_ylim()[1] > 1e3:
-                axs[1].set_ylim([0, 10])
-        else:
-            for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust):
-                plt.plot(x_values, series, 'k')
+        for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust):
+            plt.plot(x_values, series, 'k')
 
         path.format_plot(plt)
 
--- a/pyspinw/hamiltonian.py
+++ b/pyspinw/hamiltonian.py
@@ -211,7 +211,7 @@ class Hamiltonian(SPWSerialisable):
                 unique_id_to_index[input_coupling.site_1._unique_id],
                 unique_id_to_index[input_coupling.site_2._unique_id],
                 np.array(input_coupling.coupling_matrix, **rust_kw),
-                input_coupling.cell_offset.vector
+                input_coupling.cell_offset.vector.astype('double')
             )
 
             couplings.append(coupling)
@@ -222,7 +222,7 @@ class Hamiltonian(SPWSerialisable):
                 unique_id_to_index[input_coupling.site_2._unique_id],
                 unique_id_to_index[input_coupling.site_1._unique_id],
                 np.array(input_coupling.coupling_matrix.T, **rust_kw),
-                -input_coupling.cell_offset.vector
+                -input_coupling.cell_offset.vector.astype('double')
             )
 
             couplings.append(coupling)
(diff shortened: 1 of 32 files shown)


src/spinwave.rs
```diff
@@ -422,11 +422,15 @@ fn spinwave_single_q(
     // note the `faer` solver is in-place so calculates it directly on the variable `T`
     // (the input T is initially the righthand side of the equation U sqrt(E))
     let mut T = eigvecs * sqrt_E.as_diagonal();
-    solve_upper_triangular_in_place(sqrt_hamiltonian.adjoint().as_ref(), T.as_mut(), Par::Seq);
+    let mut kk = sqrt_hamiltonian.clone();
```
lucas-wilkins: Two questions here. Is there a more explicit name we could give `kk`, if not, maybe a comment? Is the clone necessary?

pyspinw/hamiltonian.py
```diff
@@ -223,7 +222,7 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
                 unique_id_to_index[input_coupling.site_2._unique_id],
                 unique_id_to_index[input_coupling.site_1._unique_id],
                 np.array(input_coupling.coupling_matrix.T, **rust_kw),
-                -input_coupling.vector(expanded.structure.unit_cell)
+                -input_coupling.cell_offset.vector.astype('double')
```
lucas-wilkins, line 241: This has changed from cartesian to LU coordinates, were you aware of this? is this the bugfix?

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
+                                 return_intensities: bool = False) -> list[np.ndarray]:
         """ Return energies as series corresponding to q, sorted by energy """
-        energy, _ = self.energies_and_intensities(path.q_points(), field=field, use_rust=use_rust)
+        energy, intensities = self.energies_and_intensities(path.q_points(), field=field, use_rust=use_rust)
 
         energy = np.array(energy)
 
         # Sort the energies
-        energy = np.sort(energy.real, axis=1)
+        idx_en = np.argsort(energy.real, axis=1)
```
lucas-wilkins: Lazy name, `energy_indices` ;)

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
     def sorted_positive_energies(self,
                                  path: Path,
                                  field: ArrayLike | None = None,
-                                 use_rust: bool = True) -> list[np.ndarray]:
+                                 use_rust: bool = True,
+                                 return_intensities: bool = False) -> list[np.ndarray]:
```
lucas-wilkins: The function no longer does what it says it does. We should probably refactor this into multiple, well named functions.

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
         # return the top half (positive)
         n_energies = energy.shape[1]
+
+        energy = np.take_along_axis(energy, idx_en, axis=1)
         energies = [energy[:, n_energies - i - 1] for i in range(n_energies//2)]
 
+        if return_intensities:
+            intensities = np.take_along_axis(np.array(intensities), idx_en, axis=1)
+            intensities = [intensities[:, n_energies - i - 1] for i in range(n_energies//2)]
+            return zip(energies, intensities)
```
lucas-wilkins: `zip` returns a zip object since python 3.something, and they don't work the same as a list, it should return something that is less likely to cause confusion.

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
-        energy = np.sort(energy.real, axis=1)
+        idx_en = np.argsort(energy.real, axis=1)
 
         # return the top half (positive)
         n_energies = energy.shape[1]
+
+        energy = np.take_along_axis(energy, idx_en, axis=1)
         energies = [energy[:, n_energies - i - 1] for i in range(n_energies//2)]
 
+        if return_intensities:
```
lucas-wilkins: The type of output should be fixed. Although we (currently) make exceptions for optimising in the rust part, having polymorphism on the output type is not good, especially for exposed code. Even in the rust code, the intention is to have a single return type with optional fields in the future.

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
                     field: ArrayLike | None = None,
                     show: bool=True,
                     new_figure: bool=True,
-                    use_rust: bool=True):
+                    use_rust: bool=True,
+                    show_intensities=False):
         """ Create a spaghetti diagram """
-        if new_figure:
+        if new_figure and show_intensities:
+            fg, axs = plt.subplots(2, 1)
```
lucas-wilkins: Lazy shortening of an already brief convention. `fig` is the convention for matplotlib stuff

lucas-wilkins: also `axes` is better, would cause collision if there was an `import *` ... but that's one reason we don't `import *`

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
+        elif show_intensities:
+            fg = plt.gcf()
+            axs = fg.get_axes()
+            for ii in range(len(axs), 2):
```
lucas-wilkins: Opposite comment here, what's wrong with `i`?

pyspinw/hamiltonian.py
```diff
@@ -253,36 +252,57 @@ def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | Non
...
-
         x_values = path.x_values()
 
-        for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust):
-            plt.plot(x_values, series, 'k')
+        if show_intensities:
+            for series in self.sorted_positive_energies(path, field=field, use_rust=use_rust, return_intensities=True):
+                axs[0].plot(x_values, series[0], 'k')
+                axs[1].plot(x_values, series[1])
+            if axs[1].get_ylim()[1] > 1e3:
```
lucas-wilkins: I get why you've done this, but maybe there's a better way?

pyspinw/lattice_distances.py
```diff
@@ -116,10 +117,10 @@ def get_cell_offsets_containing_bounding_box(
 
     limits = np.ceil(box_sizes)
 
-    # Get bounds of the form [-(l+1), l+1], note that arange is not inclusive
-    i_values = np.arange(-(limits[0]+1), limits[0]+2)
-    j_values = np.arange(-(limits[1]+1), limits[1]+2)
-    k_values = np.arange(-(limits[2]+1), limits[2]+2)
+    # Get bounds of the form [0, l+1], note that arange is not inclusive
+    i_values = np.arange(0, limits[0]+2)
```
lucas-wilkins: I've been thinking about making this change, and there is an issue about it #98 - I think its probably the in the right direction for correct way of handling it, but I think it chops off too much. I was unsure, which is why I had not done it myself.

I think what it probably needs to be is a restriction to just one side of an arbitrarily chosen plane. Assuming this plane is chosen to be aligned with two principle axes, the zeros also need to be pruned along some of the directions. Basically so that no points are related by reflection through the origin, except for 0,0,0.

pyspinw/symmetry/supercell.py
```diff
@@ -283,7 +283,9 @@ def _deserialise(json: dict, context: SPWDeserialisationContext):
             expected_names = ", ".join([f"'{key}'" for key in supercell_types])
             raise SPWSerialisationError(f"Expected transform type to be one of {expected_names}") from ke
 
-
+    @property
+    def scaling(self):
```
lucas-wilkins, line 287: Docstring, linter will force this anyway.

lucas-wilkins (changes requested):
There's some more work to be done on how the q points scale with the supercell, and I'm not 100% sure things will work right treating as you have here. I think we need a layer of converting q in and out of RLU, or things will get very confusing. Not saying that should be done here, in fact, I have previously made an issue for this #102 .

lucas-wilkins:
I'm still very skeptical about this. Fundamentally, I think the calculation should be invariant to the choice of unit cell, and I don't think this solution achieves that, even after correcting for the supercell scaling.

pyspinw/hamiltonian.py
mducle: > is this the bugfix?

Yes. I think that we need this in lattice units rather than Cartesian... I'm not 100% sure but in the original code many of the examples fail with nonsensical energies...

pyspinw/hamiltonian.py
mducle: I don't like variables with single letters because they're hard to search for and for `i` and`j` could be confused with the unit imaginary number.

pyspinw/hamiltonian.py
mducle: Not sure... what do you suggest?

pyspinw/lattice_distances.py
mducle: Actually, this code change is wrong... it's actually a bug in the coupling expansion (missing factor 1/2) which this seemed to fix for a particular case (antiferromagnetic chain).

mducle:
@lucas-wilkins I think this is ready for review... There is still a scale factor discrepancy in the intensity calculation but there's so much in this already I want to merge it before doing another PR for that fix (which I would need to investigate in more detail but don't have much time this week).

src/spinwave.rs
```diff
@@ -151,7 +151,10 @@ fn calc_sqrt_hamiltonian(
     let A_conj_minus_C: Mat<C64> = A.adjoint() - C;
     let B_adj = B.adjoint().to_owned();
 
-    let hamiltonian: Mat<C64> = block_matrix(&A_minus_C, &B, &B_adj, &A_conj_minus_C);
+    // Adds a small delta to diagonal to ensure we don't have an exact degeneracies
+    let diag_delta = Mat::<C64>::from_fn(A.nrows() * 2, A.ncols() * 2,
+        |i,j| if i == j { C64::from((i as f64)*1e-12) } else { C64::ZERO } );
+    let hamiltonian: Mat<C64> = block_matrix(&A_minus_C, &B, &B_adj, &A_conj_minus_C) + diag_delta;
```
alexhroom: how about instead of initialising a whole extra $2n \times 2n$ array for every `q`, you add minus delta to `C` in the q-independent code so $A - C$ and $A^* - C$ become $A - C + \delta$ ? I'd note also that it's generally faster to loop over `1..n` and add to each diagonal value individually, or iterate over the diagonal with

```rust
array.diagonal_mut().column_vector_mut().iter_mut().for_each(|x| *x += C64::from(1e-12) );
```

as you do later on for `T`.

src/spinwave.rs
```diff
@@ -337,10 +343,11 @@ fn spinwave_single_q(
     // pair of sites i, j; to calculate these efficiently we calculate
     // exp(i q r_i) for each site i and then the outer product of this with its conjugate
     // gives us the full matrix of phase factors.
+    let J2PI = 2. * J * PI;
     let phase_factors = Col::<C64>::from_iter(
         positions
             .iter()
-            .map(|r_i| (J * (q.transpose() * r_i)).exp()),
+            .map(|r_i| (J2PI * (q.transpose() * r_i)).exp()),
```
alexhroom: as far as I can tell this is the only place that we use `J2PI`; why not just inline 2. * J * PI in the map?
or alternatively for performance, define `J2PI` as a const rather than a variable that gets initialised separately for every single q

src/spinwave.rs
```diff
@@ -421,11 +424,13 @@ fn spinwave_single_q(
...
-    solve_lower_triangular_in_place(sqrt_hamiltonian.as_ref(), T.as_mut(), Par::Seq);
+    solve_upper_triangular_in_place(sqrt_hamiltonian.adjoint().as_ref(), T.as_mut(), Par::Seq);
 
-    // T is NaN if there are zero eigenvalues; set to zeroes
+    // T is NaN if sqrt_hamiltonian is singular; add a delta to the diagonal to avoid this
     if T.has_nan() {
-        T = Mat::<C64>::zeros(2 * n_sites, 2 * n_sites);
+        sqrt_hamiltonian.diagonal_mut().column_vector_mut().iter_mut().for_each(|x| *x += C64::from(1e-7) );
+        T = eigvecs * sqrt_E.as_diagonal();
+        solve_upper_triangular_in_place(sqrt_hamiltonian.adjoint().as_ref(), T.as_mut(), Par::Seq);
```
alexhroom, line 433: is `sqrt_hamiltonian` ever singular now we add the delta to it? is there no better way to tell whether `sqrt_hamiltonian` is singular other than performing the backward substitution?

src/lib.rs
```diff
@@ -183,7 +203,12 @@ pub fn spinwave_calculation_Sab<'py>(
         .map(faer_ext::IntoFaer::into_faer)
         .collect();
 
-    let results = calc_spinwave(r, magnitudes, q_vectors.clone(), c, p, field, true);
+    let to_cart: Mat<f64> = match rlu_to_cart {
+        Some(f) => f.into_faer().to_owned(),
+        None => Mat::<f64>::identity(3, 3),
+    };
```
alexhroom: this is pretty small and inconsequential, but personally, I would change `to_cart` into an `Option<MatRef<f64>>` rather than pass an identity matrix around. Then it can directly use the matrix memory from Python (from the `PyReadonlyArray2` object) rather than making a copy, and also avoid a trivial matrix multiplication when `rlu_to_cart` is `None`. so something like
```rust
let to_cart: Option<MatRef<f64>> = rlu_to_cart.map(|f| f.into_faer())
```
which will be `Some(ref)` where `ref` is a `MatRef` to the Python matrix if `rlu_to_cart` is given, or `None` otherwise ([see Option.map](https://doc.rust-lang.org/std/option/enum.Option.html#method.map)). Then have the coordinate conversion match the option type and only perform the transform if there is a transform to do!

src/lib.rs
```diff
@@ -146,7 +160,12 @@ pub fn spinwave_calculation<'py>(
         .map(faer_ext::IntoFaer::into_faer)
         .collect();
 
-    let results = calc_spinwave(r, magnitudes, q_vectors.clone(), c, p, field, false);
+    let to_cart: Mat<f64> = match rlu_to_cart {
+        Some(f) => f.into_faer().to_owned(),
+        None => Mat::<f64>::identity(3, 3),
+    };
+
```
alexhroom: see comment on the other occurrence of this type conversion

alexhroom (changes requested):
just reviewing the Rust as you asked - mostly quality code, just a few little bits here and there for performance

src/spinwave.rs
mducle: I think that it could still be singular - in all the cases I've tested adding a delta was enough to make it non-singular but I haven't tested many cases!

I guess we could compute the condition number and if it was high (some high limit to be defined) then we assume it is singular.

mducle: So computing the condition number requires a `svd` decomposition so is probably just as expensive as solving it. The other option is to compute the determinant but that needs lu decomposition which might also be just as expensive... Given this, I'd rather leave this as is...

alexhroom: fair enough - just wanted to double-check

alexhroom (approved):
happy with the Rust now!

PR #115, opened 2026-02-25
Update interface functions

mducle:
* Change `sites()` to `generate_sites()` and remove transformation functionality (move to `generate_structure` below)
* Add `generate_structure()` function to output a `Structure` object from a `UnitCell` and list of sites/moments. This supports different units (`xyz` and `lu` for both moments and positions separately as well as allow to specify spin lengths (magnitudes).
* Add `generate_helical_structure()` to act like `genmagstr('mode','helical',...` in Matlab.
* Add `bond` option in ~~`couplings()`~~ `generate_exchanges` to allow users to specify the bond index rather than a distance range
* Updated `omegasum` function to output equal number of modes to prevent spikes in the energy (still get cross-over jumps in intensity though due to energy-sorting).
* Add `egrid` function to calculate energy-width-convolved spectrum and changed `spaghetti_plot` to use this renaming old function to `spaghetti_plot_dual`.
* Add system tests comparing to Matlab output.


mducle:
@lucas-wilkins - tests and examples to follow - just wanted to get your opinion on if this is ok structure wise, particular about the separate units for positions and moments and need for a `UnitCell` to convert if either of these are not the default.

The alternative option is to _not_ have any units in `generate_sites` but only have a different set of `generate_structure`s interface functions which would require a `UnitCell` as input (like `generate_helical_structure` implemented here).

lucas-wilkins:
yeah, that seems reasonable.

I think we should maybe change everything to `generate_*`, not just the sites

If you look at `propagation_vectors` that has some extra parameters (namely phase) that maybe we want to include as well.

I think I prefer avoiding using complex numbers for the moment, and using rotation matrices, but just a personal preference.

Enum is good, but maybe it should live somewhere else eventually.

mducle:
@lucas-wilkins @henrikjacobsenfys - I think this PR is ready for review.

~~To actually run any full calculation, you need to also have the changes from the `sab_deb` branch too. I would recommend to do this `git checkout main && git merge sab_deb && git merge interfaces_generate_sites`~~

[Edit: updated main and rebased branch so it should work if you just checkout this branch].

Then you can download this [gist script](https://gist.github.com/mducle/d5eb536bc7f1b2128abacdd1e18cc17a) and run it, comparing the output with this [matlab gist](https://gist.github.com/mducle/72ba28fd46010def8e8eb3b9f4012dab).

pyspinw/couplinggroup.py
```diff
@@ -103,19 +105,30 @@ def __init__(self,
 
     def couplings(self, sites: list[LatticeSite], unit_cell: UnitCell):
         """ Get the couplings defined by this groups given the sites and symmetry provided """
+        if self.bond > 0:
```
henrikjacobsenfys, line 110: Hmmm, if people are doing long distance interactions like RKKY then this will give problems. I don't know if responsibilities become too muddled, but can self.max_distance be used here if it exists?

pyspinw/couplinggroup.py
```diff
@@ -103,19 +105,30 @@ def __init__(self,
...
+
         abstract_couplings = batch_couplings(
             sites=sites,
             unit_cell=unit_cell,
-            max_distance=self.max_distance,
+            max_distance=distances[1],
             naming_pattern=default_naming_pattern if self.naming_pattern is None else self.naming_pattern,
             type_symbol=self.coupling_type.short_string)
 
+        if self.bond > 0:
```
henrikjacobsenfys, line 123: I seem to recall that bonds can have the same distance and not be the same bond. It'probably s a lot more work, but can the unique bonds be determined by symmetry?

pyspinw/hamiltonian.py
```diff
@@ -176,11 +177,21 @@ def print_summary(self):
...
-    def energies_and_intensities(self, q_vectors: np.ndarray, field: ArrayLike | None = None, use_rust: bool=True):
-        """Calculate the energy levels of the system for the given q-vectors."""
-        #
-        # Set up choice of calculation
-        #
+    def energies_and_intensities(self,
+                                 q_vectors: np.ndarray,
+                                 field: ArrayLike | None = None,
+                                 use_rust: bool=True,
+                                 intensity_unit: IntensityUnits | str='cell'):
```
henrikjacobsenfys: I might be overcomplicating things, but in my mind there are two separate things here: whether to normalise to unit cel or atoms, and whether to use mbarn/sr/meV/ (unit cell or atom) or the other units (I forgot what they are).

pyspinw/interface.py
```diff
@@ -292,10 +382,11 @@ def couplings(sites: list[LatticeSite],
...
 
-    :param sites: *required* List of sites to make couplings between
-    :param unit_cell: *required* Unit cell (needed for working out the Cartesian coordinates of sites)
-    :param coupling_type: *required* Type of coupling
-    :param max_distance: *required* Maximum Cartesian distance at which couplings are made
+    :param sites: *required* List of sites to make couplings between or a Structure object
+    :param unit_cell: Unit cell (needed if first argument is a list of sites)
+    :param coupling_type: Type of coupling (defaults to HeisenbergCoupling)
+    :param bond: The bond index (If this is given the _distance parameters are ignored)
+    :param max_distance: Maximum Cartesian distance at which couplings are made
```
henrikjacobsenfys: May be good to inform users of the unit of `min_distance` and `max_distance` here

henrikjacobsenfys (commented):
Looks good, just a few questions and comments. I haven't tried the code yet

pyspinw/couplinggroup.py
mducle: So `bond` here is meant to be a scalar - every coupling covered in this `CouplingGroup` has the same exchange matrix. It's probably not very efficient to regenerate the list of all bonds up to some distance to cover the $`n^{\mathrm{th}}`$ bond each time but it's what is implemented here... We can have another interface function for an exchange function type $`J(\mathbf{r})`$ for interactions like RKKY (or dipolar).

pyspinw/couplinggroup.py
mducle: Yeah, there should be some machinery already implemented to do this - @lucas-wilkins could you have a look at this?

pyspinw/hamiltonian.py
mducle: The current units are "spin-length-squared per unit cell" - if you look at eq 35/42/43 of the paper it's just the Fourier transform of the spin-spin correlation function... Since spin lengths are unit-less the units would technically be "per unit cell"... 

Anyway, `IntensityUnits` is meant to be an enum with three values `PERCELL`, `PERATOM` and `BARN` ([code here](https://github.com/SpinW/pySpinW/pull/115/changes#diff-ee2bb52297740e1f4e6f8a06b1d88f06a28f9c2c7fad23f4715fe06a2ec03012R14-R16)). You can see below that `BARN` is not implemented yet, but will be.

`BARN` in this case is per atom but if you want it to be `barn/sr/meV/unit-cell` there could be another option `BARNPERCELL` too. (I didn't think that barn/sr/meV/unit-cell made sense...)

pyspinw/couplinggroup.py
henrikjacobsenfys: I assume you mean int :)

I was more thinking if users for some reason want an exchange that is longer than the max you give then they currently have no way of getting it. Then again, RKKY is the only example I can think of with such long range interactions, so maybe it's better to make another interface function as you say.

pyspinw/hamiltonian.py
henrikjacobsenfys: My point was that there are 4 options: (`s-l-spuc` (name up for debate) or `BARN`) and (`PERCELL` or `PERATOM`). They're not mutually exclusive.

Some users might also ask to plot things per formula unit, but a line has to be drawn somewhere..

pyspinw/couplinggroup.py
```diff
@@ -103,19 +105,30 @@ def __init__(self,
...
+            bond_distances = np.sort(np.unique(np.round([c.distance for c in abstract_couplings], decimals=6)))
```
RichardWaiteSTFC: This seems reasonable to round to 6dp - but perhaps we could give the user the option to specify this?

pyspinw/couplinggroup.py
henrikjacobsenfys: Never mind, I apparently can't read code. It works just fine

pyspinw/hamiltonian.py
```diff
@@ -60,6 +61,19 @@ def omegasum(energy: ArrayLike, intensity: ArrayLike, tol: float=1e-5, zeroint:
     int_out = np.delete(int_out, np.where(nanmodes == en_out.shape[0])[0], axis=1)
     return en_out, int_out
 
+
+def egrid(energy: ArrayLike, intensity: ArrayLike, evect: ArrayLike, dE: float | Callable):
+    """Bins a set of energy/intensity into a spectrum with Gaussian broadening"""
+    esigma = dE(evect) if isinstance(dE, Callable) else np.ones(evect.shape) * dE
+    esigma /= 2 * np.sqrt(2* np.log(2))  # Convert from FWHM
+    spec = np.zeros((energy.shape[0], len(evect)))
+    ew_norm = np.array(list(np.diff(evect)) + [evect[-1] - evect[-2]]) / (np.sqrt(2 * np.pi) * esigma)
```
RichardWaiteSTFC: I think if possible it would be best to avoid converting to list and back to array - there is an option in `np.diff` to append an element - is this equivalent?
`ew_norm = np.diff(evect, append=evect[-1] + (evect[-1] - evect[-2])) / (np.sqrt(2 * np.pi) * esigma)`
I also wonder whether it's possible to do some of this with `scipy.stats.gaussian_kde` with the intensities added as weights? It's might not work as it's for making pdfs, there might be some normalization needing undoing etc.

PR #117, opened 2026-03-10
Fix coupling generation

### LINKED ISSUE #98: Check Coupling Groups for duplicates

lucas-wilkins:
This *should* fix the behaviour of the coupling generation.

I have removed the hack.

I've added a couple of tests, but more would be good. @mducle, if you could try the things you were concerned about, that would be good.

Fixes #98 - probably


pyspinw/batch_couplings.py
```diff
@@ -4,6 +4,7 @@
 
 import re
 from dataclasses import dataclass
+from http.cookiejar import is_HDN
```
mducle: Unneeded import?

tests/test_coupling_generation.py
```diff
@@ -0,0 +1,41 @@
...
+    assert len(couplings) == 2
+
+def test_heisenberg_antiferromagnet_supercell():
+    sites = [LatticeSite(0, 0, 0, 0, 0, 1, name="X")]
+
+    couplings = generate_exchanges(sites, UnitCell(1, 1, 1), HeisenbergCoupling, 1.1,
+                                   direction_filter=BiDirectionFilter([1, 0, 0]), j=1)
+
+    assert len(couplings) == 1
+
```
mducle, line 41: I'd like to see a couple of more tests with different geometries, e.g.

```python
sites = [LatticeSite(0, 0, 0, 0, 0, 1, name="X")]
# Hexagonal
couplings = generate_exchanges(sites, UnitCell(1, 1, 2, gamma=120), HeisenbergCoupling, 1.1, j=1)
assert len(couplings) == 3
# Cubic - nearest neighbour
couplings = generate_exchanges(sites, UnitCell(1, 1, 1), HeisenbergCoupling, 1.1, j=1)
assert len(couplings) == 3
# Cubic - second neighbour (face diagonal * 3 faces)
couplings = generate_exchanges(sites, UnitCell(1, 1, 1), HeisenbergCoupling, 1.5, 1.1, j=1)
assert len(couplings) == 6
# Cubic - third neighbour (body diagonal)
couplings = generate_exchanges(sites, UnitCell(1, 1, 1), HeisenbergCoupling, 1.8, 1.5, j=1)
assert len(couplings) == 4
```

mducle (changes requested):
Looks good, and does fix the duplicate couplings problem. There's just an extraneous `import` to remove and a few more tests to add and it's good to go.

mducle (approved):

PR #118, opened 2026-03-10
Add coverage PR comment to gh-actions

mducle:
Adds a comment to a PR summarising test code coverage.

Note that currently the Python `spinwave.py` code is not checked because we run it in separate subprocesses using `concurrent.futures.ParallelPoolExecutor`. This PR might add an option to disable this paralellisation.

This is a very basic addition. There are other github actions which have more features such as [MishaKav](https://github.com/MishaKav/pytest-coverage-comment) or [py-cov-action](https://github.com/py-cov-action/python-coverage-comment-action). In particular they create a new branch which uploads the html coverage reports so you can browse it. There are also external services like coveralls.io or codecov.io but I didn't want to use them as it requires additional permissions and will not work for forked PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,9 +37,25 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
         pip install -r requirements-test.txt
-    - name: Run tests with pyTest
+    - name: Run tests with pyTest (non-Linux)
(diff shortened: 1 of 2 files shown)


henrikjacobsenfys (approved):
Looks good to me. Also seems like some classes might need a few more tests ;)

PR #122, opened 2026-03-12
Rotating Frame

### LINKED ISSUE #119: Implement rotating frame calculation
One key unique capability of SpinW is the use of a rotating coordinate frame for the spins for helical structures. Toth and Lake showed that for magnetic structures describable in this way, that the magnon modes can be split into three sets of branches which can be computed more cheaply using the structural (crystal) unit cell rather than a larger magnetic supercell.

We should implement this option in PySpinW, but with additional guard rails to ensure that it only applies to magnetic structures (mostly helices) which can be described by a rotating coordinate frame. (In the Matlab code any "incommensurate" structure triggers this calculation mode even for structures such as amplitude modulated structures which cannot be described by a rotating coordinate frame).

mducle:
Add a rotating frame (aka "incommensurate") calculation mode in both Python and Rust.

Also, created a new `RotationSupercell` to specifically address helical structures, because the $`S(\mathbf{Q},\omega)`$ calculation requires a rotation axis direction which is not otherwise stored anywhere (it's now a property of `RotationSupercell`. This also enables changing the `LatticeSite` moment property back to being real.

Fix #119

### DIFF
--- a/src/spinwave.rs
+++ b/src/spinwave.rs
@@ -1,8 +1,8 @@
 use std::f64::consts::PI;
 
 use faer::linalg::triangular_solve::solve_upper_triangular_in_place;
-use faer::{unzip, zip, Col, ColRef, Mat, MatRef, Par, Side, perm, mat};
-use faer::mat::{AsMatRef, AsMatMut};
+use faer::mat::{AsMatMut, AsMatRef};
+use faer::{mat, perm, unzip, zip, Col, ColRef, Mat, MatRef, Par, Side};
 use indicatif::ParallelProgressIterator;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 
@@ -90,7 +90,11 @@ fn calc_q_independent(
     C *= 2.;
 
     // Adds a small delta to diagonal to ensure we don't have an exact degeneracies
-    C.diagonal_mut().column_vector_mut().iter_mut().enumerate().for_each(|(i, x)| *x -= C64::from((i as f64)*1e-12) );
+    C.diagonal_mut()
+        .column_vector_mut()
+        .iter_mut()
+        .enumerate()
+        .for_each(|(i, x)| *x -= C64::from((i as f64) * 1e-12));
 
     // if an external magnetic field is provided, calculate the Az matrix
     // which is added to the diagonal of A in the Hamiltonian
@@ -188,9 +192,14 @@ fn calc_sqrt_hamiltonian(
             zip!(&mut sqrt_d, d).for_each(|unzip!(sqd, v)| *sqd = v.sqrt());
             // need to apply permutations: in Python scipy does this for you
             let mut shm = Mat::<C64>::zeros(l.nrows(), l.ncols());
-            perm::permute_cols(shm.as_mat_mut(), (ldl.P().inverse() * l * sqrt_d.as_diagonal()).as_mat_ref(), ldl.P().inverse());
+            perm::permute_cols(
+                shm.as_mat_mut(),
+                (ldl.P().inverse() * l * sqrt_d.as_diagonal()).as_mat_ref(),
+                ldl.P().inverse(),
+            );
             shm
-        }} 
+        }
+    }
 }
 
 /// Calculate energies (eigenvalues of the Hamiltonian) for a set of q-vectors.
@@ -310,13 +319,19 @@ pub fn calc_spinwave(
             let km: Col<f64> = rot_comps[0].to_owned();
             let n = Col::<C64>::from_iter(rot_comps[1].iter().map(|f| C64::new(*f, 0.)));
             // Computes the rotation matrices in Toth & Lake eq 39
-            let nx = mat![[C64::ZERO, -n[2], n[1]], [n[2], C64::ZERO, -n[0]], [-n[1], n[0], C64::ZERO]];
+            let nx = mat![
+                [C64::ZERO, -n[2], n[1]],
+                [n[2], C64::ZERO, -n[0]],
+                [-n[1], n[0], C64::ZERO]
+            ];
             let R2 = n.clone() * n.transpose();
             let R1 = (Mat::<C64>::identity(3, 3) - (nx.as_ref() * SCALAR_J) - R2.as_ref()) / 2.;
             new_couplings = Vec::from_iter(couplings.iter().map(|c| {
                 let phi = 2. * PI * km.transpose() * c.inter_site_vector.as_ref();
                 // R is the Rodrigues rotation matrix
-                let R = Mat::<C64>::identity(3, 3) * phi.cos() + &nx * phi.sin() + (1. - phi.cos()) * &R2;
+                let R = Mat::<C64>::identity(3, 3) * phi.cos()
+                    + &nx * phi.sin()
+                    + (1. - phi.cos()) * &R2;
                 Coupling {
                     index1: c.index1,
                     index2: c.index2,
@@ -324,9 +339,9 @@ pub fn calc_spinwave(
                     inter_site_vector: c.inter_site_vector.clone(),
                 }
             }));
-            couplings = Vec::from_iter(new_couplings.iter().map(|f| f));
+            couplings = Vec::from_iter(new_couplings.iter());
             Some(RotatingFrameComponents { km, nx, R1, R2 })
-        },
+        }
         _ => None,
     };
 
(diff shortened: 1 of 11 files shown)


src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
+    // Need a second vector with mutable elements from which we can borrow due to Pyo3 limits.
+    let new_couplings: Vec<Coupling>;
+    let rotating_components: Option<RotatingFrameComponents> = match rotating_frame {
+        Some(rot_comps) => {
+            let km: Col<f64> = rot_comps[0].to_owned();
+            let n = Col::<C64>::from_iter(rot_comps[1].iter().map(|f| C64::new(*f, 0.)));
+            // Computes the rotation matrices in Toth & Lake eq 39
+            let iNx = mat![[C0, -n[2] * J, n[1] * J], [n[2] * J, C0, -n[0] * J], [-n[1] * J, n[0] * J, C0]];
+            let nx = mat![[C0, -n[2], n[1]], [n[2], C0, -n[0]], [-n[1], n[0], C0]];
+            let R2 = n.clone().as_mat() * n.as_mat().transpose();
```
alexhroom: I don't think the `as_mat` conversion is necessary to do a vector product - elsewhere in the code we just use `vec.clone() * vec.transpose()` (see e.g. the spin coefficients, or phi a few lines down). unless I've missed something

src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
+    let mut couplings = in_couplings;
+    // Need a second vector with mutable elements from which we can borrow due to Pyo3 limits.
+    let new_couplings: Vec<Coupling>;
+    let rotating_components: Option<RotatingFrameComponents> = match rotating_frame {
+        Some(rot_comps) => {
+            let km: Col<f64> = rot_comps[0].to_owned();
+            let n = Col::<C64>::from_iter(rot_comps[1].iter().map(|f| C64::new(*f, 0.)));
+            // Computes the rotation matrices in Toth & Lake eq 39
+            let iNx = mat![[C0, -n[2] * J, n[1] * J], [n[2] * J, C0, -n[0] * J], [-n[1] * J, n[0] * J, C0]];
+            let nx = mat![[C0, -n[2], n[1]], [n[2], C0, -n[0]], [-n[1], n[0], C0]];
```
alexhroom: why not define `nx` and then let `iNx` be `nx.clone() * J`?

you may have to use the const `SCALAR_J` from `crate::constants`

src/spinwave.rs
```diff
@@ -13,6 +13,7 @@ use crate::{Coupling, MagneticField, C64};
 /// Minimum energy that isn't just set for zero (in meV)
 const ZERO_ENERGY_TOL: f64 = 1e-12;
 const J2PI: C64 = C64::new(0., 2. * PI);
+const C0: C64 = C64::new(0., 0.);
```
alexhroom, line 16: `num_complex` already implements this constant as `C64::ZERO`

src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
+                &rotating_components,
                 save_Sab,
             )
         })
         .collect()
 }
 
+/// Caculate energies and intensities for a triplet of q-vectors (rotating frame)
+/// For non-rotating frame calculations, just calls spinwave_single_q
+fn spinwave_triq(
```
alexhroom: lengthen to `spinwave_triplet_q` so it's clearer what this is?

src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
+                    Some(vec![Sm.sab.unwrap(), S0.sab.unwrap(), Sp.sab.unwrap()].concat())
+                },
+                false => None,
+            };
+            SpinwaveResult {
+                energies: vec![Sm.energies, S0.energies, Sp.energies].concat(),
+                sab: sab,
+                intensities: vec![Sm.intensities, S0.intensities, Sp.intensities].concat(),
+            }
+        },
```
alexhroom: i think there's a nicer way to do this... [[[[[[[[[LOOK OVER BEFORE SUBMITTING REVIEW]]]]]]]]]

also we can paralellise these three so they all run simultaneously

src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
                 n_sites,
                 &couplings,
                 &positions,
                 rlu_to_cart,
+                &rotating_components,
                 save_Sab,
             )
         })
         .collect()
 }
```
alexhroom: I know you wanted to avoid having the two calls for rotating or not rotating, but I think in this case it'd make it far clearer and more intuitive with what's going on - you already have a branch based on whether or not you're rotating at the `spinwave_triq` level so you may as well have it here instead e.g. if it was

```rust
if rotating_frame {
// map with spinwave_triq
} else {
// map with spinwave_single_q
}
```

and the `match rot` was removed from `spinwave_triq`

src/spinwave.rs
```diff
@@ -279,34 +293,98 @@ pub fn calc_spinwave(
...
+                let R = Mat::<C64>::identity(3, 3) * phi.cos() + &nx * phi.sin() + (1. - phi.cos()) * &R2;
+                Coupling {
+                    index1: c.index1,
+                    index2: c.index2,
+                    matrix: (c.matrix.as_ref() * R.as_ref() + R.as_ref() * c.matrix.as_ref()) / 2.,
+                    inter_site_vector: c.inter_site_vector.clone(),
+                }
+            }));
+            couplings = Vec::from_iter(new_couplings.iter().map(|f| f));
+            Some(RotatingFrameComponents { km: km, nx: nx, R1: R1, R2: R2 })
```
alexhroom: neat little Rust feature is that if your variable names match the field names, you don't actually need to repeat them:

```suggestion
            Some(RotatingFrameComponents { km, nx, R1, R2 })
```
would be valid

alexhroom (changes requested):
mostly just some tidying to do!

src/spinwave.rs
mducle: Yes, `SCALAR_J` works... I tried `nx * J` first but that didn't work.

src/spinwave.rs
```diff
@@ -279,32 +292,94 @@ pub fn calc_spinwave(
...
+                // R is the Rodrigues rotation matrix
+                let R = Mat::<C64>::identity(3, 3) * phi.cos() + &nx * phi.sin() + (1. - phi.cos()) * &R2;
+                Coupling {
+                    index1: c.index1,
+                    index2: c.index2,
+                    matrix: (c.matrix.as_ref() * R.as_ref() + R.as_ref() * c.matrix.as_ref()) / 2.,
+                    inter_site_vector: c.inter_site_vector.clone(),
+                }
+            }));
+            couplings = Vec::from_iter(new_couplings.iter().map(|f| f));
```
alexhroom: what is this `map` doing? `cargo clippy` believes it is redundant

src/spinwave.rs
```diff
@@ -315,23 +390,32 @@ pub fn calc_spinwave(
...
     rlu_to_cart: Option<MatRef<f64>>,
+    rotating_components: &Option<RotatingFrameComponents>,
     save_Sab: bool,
+    tri_id: f64,
 ) -> SpinwaveResult {
     let z = &q_independent_components.z;
     let spin_coefficients = &q_independent_components.spin_coefficients;
 
+    if let Some(rotcomp) = rotating_components {
+        q = q + tri_id * rotcomp.km.as_ref();
```
alexhroom:
```suggestion
        q += tri_id * rotcomp.km.as_ref();
```

src/spinwave.rs
```diff
@@ -452,6 +536,25 @@ fn spinwave_single_q(
...
+                                                    + (&rotcomp.R2 * &*m * &R22I)));
+        // Apply the rotation transformation (eq 40)
+        match tri_id {
+            n if n < 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R1.conjugate()),
+            n if n > 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R1),
+            n if n == 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R2),
+            _ => panic!(),
+        }
+        // Convert q back for qperp calculation
+        q = q - tri_id * rotcomp.km.as_ref();
```
alexhroom:
```suggestion
        q -= tri_id * rotcomp.km.as_ref();
```

src/spinwave.rs
```diff
@@ -452,6 +536,25 @@ fn spinwave_single_q(
...
+        let R2I = &rotcomp.R2 - Mat::<C64>::identity(3, 3);
+        let R22I = (2. * &rotcomp.R2) - Mat::<C64>::identity(3, 3);
+        Sab.iter_mut().for_each(|m| *m = 0.5 * (&*m - (&rotcomp.nx * &*m * &rotcomp.nx)
+                                                    + (&R2I * &*m * &rotcomp.R2)
+                                                    + (&rotcomp.R2 * &*m * &R22I)));
+        // Apply the rotation transformation (eq 40)
+        match tri_id {
+            n if n < 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R1.conjugate()),
+            n if n > 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R1),
+            n if n == 0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R2),
```
alexhroom:
```suggestion
            0. => Sab.iter_mut().for_each(|m| *m = &*m * &rotcomp.R2),
```

alexhroom (changes requested):
sorry, just ran `cargo clippy` (linter) as a final check and it threw up some mostly trivial changes...

could I also ask you to run `cargo fmt` to format the code please?

src/spinwave.rs
mducle: yeah, it is redundant... I think I was just copying and pasting!

mducle:
@alexhroom Thanks. I also ran `cargo clippy` as well as `cargo fmt` and clippy says all our functions have to many arguments... (`calc_spinwave`, `spinwave_single_q`, `spinwave_calculation` and `spinwave_calculation_Sab`) - should we just ignore this or should I make a maintenance issue to clean up the Rust code?

alexhroom:
@mducle yeah, they've been there for a while, I think we could definitely reduce the arguments on the internal functions (in particular a few things could be moved into QIndependentComponents, like the `rlu_to_cart` matrix or the `RotatingFrameComponents` struct) 

however for the interface functions, the problem is more that we have to bundle the entire context of the system into Rust from Python. the best thing I could see doing is putting all the inputs into a struct and then passing that struct, which I don't think is any tidier than having lots of arguments...

alexhroom (approved):
happy with the Rust

lucas-wilkins (approved):

PR #124, opened 2026-03-12
97 mean field optimisation

### LINKED ISSUE #97: Mean field optimisation
Add in code to do a mean field approximation optimisation of the ground state

lucas-wilkins:
Finds the classical ground state by rotating spins under gradient descent


debug/optimisation_test_1.py
```diff
@@ -0,0 +1,56 @@
+from multiprocessing import freeze_support
+
+import numpy as np
+
+from pyspinw.calculations.energy_minimisation import ClassicalEnergyMinimisation, Free, Fixed
+from pyspinw.interface import couplings
```
mducle: I guess this was old code? (`couplings` changed to `generate_exchanges`) Also [line 26](https://github.com/SpinW/pySpinW/pull/124/changes#diff-4c54048410cece2a97cfcaf20a5cb7a4eb9b62a56c3be6989354c54f46eb27e9R26)

debug/optimisation_test_1.py
```diff
@@ -0,0 +1,56 @@
...
+        moment_history.append(minimiser.moment_data.copy())
+
+    print(minimiser.moment_data)
+
+    import matplotlib.pyplot as plt
+    ax = plt.figure().add_subplot(projection='3d')
+
+    for i in range(len(sites)):
+        xyz = np.array([moments[i,:] for moments in moment_history])
+        ax.plot(xyz[:, 0], xyz[:, 1], xyz[:, 2], "--x")
```
mducle: This line gave me an index error saying that `xyz` has shape `(101,1,3)`. If I change line 53 to `xyz = np.squeeze(np.array([moments[i,:] for moments in moment_history]))` it works ok.

docs/developers/design/007_energy_minimisation.md
```diff
@@ -0,0 +1,157 @@
...
+
+but sometimes it is a scalar, and sometimes a matrix. For efficiency we can code for either of these options, for uniformity we can multiply scalar values by the identity matrix.
+
+### Trivial supercell
+
+`TrivialSupercell`s are the identity and the derivative is also the identity. They only have one input component
+
+### Transformation supercells
+
+`TransformationSupercell`s are a linear map on one component, where $f(T) = M T$, where $M$ is matrix that depends on the which cell is selected in the supercell. This means that
```
mducle: Unfinished sentence...

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+        self.n_components = hamiltonian.structure.supercell.n_components()
+
+        self.free_sites = []
+        self.fixed_sites = []
+        self.planar_sites = []
+        self.planar_axes = []
+
+        self.is_free = np.zeros((self.n_sites,), dtype=bool)
+        self.is_fixed = np.zeros((self.n_sites,), dtype=bool)
+        self.is_planar = np.zeros((self.n_sites,), dtype=bool)
```
mducle: Out of these three, only `is_free` is used in later code...

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+                    self.is_free[site_index] = True
+
+                case FixedConstraint():
+                    self.fixed_sites.append((site_index, site.unique_id))
+                    self.is_fixed[site_index] = True
+
+                case Planar(axis):
+                    self.planar_sites.append((site_index, site.unique_id))
+                    self.planar_axes.append(axis)
+                    self.is_planar[site_index] = True
```
mducle: Using a `match` here and separate blocks of code in `jitter()`, `randomize()` and `iterate` means that the classes are only labels so (imho) you might as well just use strings or `Enum`s and just have  `MinimisationConstraintGenerator` classes with the `generate()` method (rather the than the current multiple inheritance).  

On the other hand if each `MinimisationConstraint` class could have `operate(moment)`, `randomize()`, `forces(moment, couplings, othersites)` methods - then the code in `jitter`, `randomize` and `iterate` could call these methods... and makes cleaner code there and also makes adding more constraints later on a bit easier... but this is probably going to take too long to implement...

docs/developers/design/007_energy_minimisation.md
```diff
@@ -0,0 +1,157 @@
+# Classical Energy Minimisation
+
+## Problem
+
+We will think of the Hamiltonian in form, which is defined in terms of the mathematical form, not the physical meaning.
```
henrikjacobsenfys: I don't understand this sentence

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+    """ Specify that moment should be unconstrained """
+
+    name = "Free"
+
+    def generate(self, n_constraints: int):
+        """ Generate a list of constraints """
+        return [self for _ in range(n_constraints)]
+
+
+class FixedConstraint(MinimisationConstraint):
```
henrikjacobsenfys, line 41: Shouldn't this have a generate method?

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+
+class Planar(MinimisationConstraint, MinimisationConstraintGenerator):
+    """ Specify that moment should be constrained to plane """
+
+    name = "Planar"
+
+    __match_args__ = ('axis', )
+
+    def __init__(self, axis: ArrayLike):
+        self.axis = np.array(axis, dtype=float)
```
henrikjacobsenfys, line 55: should the input be checked?

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+        self.axis /= np.sqrt(np.sum(self.axis**2))
+
+    def generate(self, n_constraints: int):
+        """ Generate a list of constraints """
+        return [self for _ in range(n_constraints)]
+
+    def __repr__(self):
+        return f"{self.name}(axis={self.axis[0]},{self.axis[1]},{self.axis[2]})"
```
henrikjacobsenfys: may want to print only a few decimals

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+        if isinstance(constraints, list):
+            if len(constraints) != n_sites:
+                raise ValueError("If constraints are a list, there should be one for each site")
+
+        elif isinstance(constraints, MinimisationConstraintGenerator):
+            constraints = constraints.generate(n_sites)
+
+        else:
+            raise TypeError("Expected constraints to be a MinimisationConstraintGenerator or "
+                            "list of MinimisationConstraints")
```
henrikjacobsenfys: maybe add what type it was?

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+        if field is None:
+            field = np.zeros((3, ))
+        else:
+            field = np.array(field)
+
+        if field.shape != (3, ):
+            raise ValueError("Expected field to be a length 3 vector")
```
henrikjacobsenfys: same here

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+
+        assert self.magnitudes.shape == (self.n_sites, self.n_components, 1)
+
+        self._site_uid_to_index = {site.unique_id: i for i, site in enumerate(self.sites)}
+
+    @staticmethod
+    def _random_orientations(rng, shape, dim=3):
+        """ Make random gaussian vectors, avoiding 0,0,0..."""
+        n = 1
+        for m in shape:
```
henrikjacobsenfys: n = np.prod(shape)

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+        #
+
+        for i in range(max_iters-1):
+            last_energy = this_energy
+            self.iterate(step_size)
+            this_energy = self.energy()
+
+            delta_energy = this_energy - last_energy
+
+            if np.abs(delta_energy) < rtol * np.abs(start_delta_energy):
```
mducle, line 368: I'm not sure that `abs` is correct here... in principle we would want each iteration to lower the energy (so `delta_energy` should be negative... the `abs` in this check would mean that a small *increase* in energy would also be treated as converged.

Another thing I'm also not sure about is the relative tolerance because that depends a lot on the initial structure - if this is very far away from the optimum then `start_delta_energy` could be quite large, and vice versa. In the first case we might not actually be converged when the tolerance is reached; in the second case the `atol` criteria would usually be reached first. So I'm not sure if `rtol` is really necessary.

pyspinw/calculations/energy_minimisation.py
```diff
@@ -0,0 +1,592 @@
...
+        # If you want a physical interpretation, this is critically damped movement, where step_size_factor is dt
+        # AKA 90s game engine physics, AKA Aristotlean physics, F = mv
+        #
+        # As we're in a coordinate system around current location, alpha = delta alpha
+
+        alpha = step_size * forces_free_alpha
+        beta = step_size * forces_free_beta
+        theta = step_size * forces_planar
+
+        if np.any(np.abs(alpha) > 1.0) or np.any(np.abs(beta) > 1.0) or np.any(np.abs(theta) > 1.0):
```
henrikjacobsenfys: Maybe these "magic numbers" should be defined in the beginning of the file with descriptive names?

mducle (changes requested):
Just a couple of things in the `debug/` script and a dangling sentence in the doc. Otherwise a couple of comments / observation on the minimiser code which maybe doesn't need changes.

One thing we should investigate later is the ability to use an adaptive step... I think there was a paper about this somewhere but I can't find it.

henrikjacobsenfys (commented):
Just a few small comments. I haven't had the time to look at the equations and their implementations, though, but looks good

mducle (approved):

PR #131, opened 2026-03-27
88 parallel python processes currently require boilerplate

### LINKED ISSUE #88: Parallel python processes currently require boilerplate
The current parallel python implementation requires scripts to be wrapped in

```
if __name__ == "__main__":
    freeze_support()

   ... [script] ...
```

This is quite annoying, perhaps there is a better way of assuring parallel processes work correctly.

lucas-wilkins:
This makes it so that we don't have to write

```
if __name__ == "__main__":
    freeze_support()

    [code here]
```

By default, this update will means the python (but **not rust**) parallisation is disabled by default **on windows**. However,
one can still use this:

```
if __name__ == "__main__":
    set_up_windows_python_parallelisation()

    [your code]
```

to enable parallisation of python on windows.

As we discussed before, this is a much better solution because

* No boilerplate unless you want python parallelisation on windows
* By default, unless running from source without compiling the rust, most users won't be using the python implementation
* Windows multiprocess parallelisation is slower for small calculations anyway, because of the overhead of making multiple instances of pyspinw

---

This also adds a `tutorials` folder, which is like the `structures` folder that I have cleaned up a bit for the standard user, and intend to build upon when doing the docs.


pyspinw/windows_parallelisation.py
```diff
@@ -0,0 +1,40 @@
...
+    """ Get an Executor appropriate for the current system """
+    if platform.system() == "Windows":
+        if not windows_python_parallelisation_enabled:
+            return ThreadPoolExecutor
+
+    return ProcessPoolExecutor
+
+
+
+def set_up_windows_python_parallelisation():
```
mducle, line 19: The only substantive change I'd like is to make this function name shorter. Maybe `setup_win_py_parallel` or `windows_parallel_on`?

mducle (changes requested):
The main thing I'd like changed is instead of having essentially duplicated files in `examples/tutorials` and `examples/structures` that the originals in `structures` could be moved to `tutorials` instead.

Also a comment about the guard function name.

mducle (approved):

PR #141, opened 2026-04-07
91 renaming of things

### LINKED ISSUE #91: Renaming of things
Some things need to be renamed, or may best be renamed. Plan is to keep this ticket open for a while, then do a refactor later for multiple potential renames.

* `mi`, `mj`, `mk` : these actually use the orthogonal cell oriented system, maybe call the coordinates in this system `r`,`s`,`t`
* Do we want to continue using `Coupling` after all the coupling names, or simplify them. Is `Exchange` better?
* `LatticeSite` is very long
* Rename `TrivialSupercell`?
* Rename `SummationSupercell` to `FourierSupercell`
* Change use of `fractional` to `lu`?

lucas-wilkins:
Addresses #91 

**Coupling -> Exchange**

I have renamed coupling to exchange in the user facing side, and in (I think) all of the relevant internal places too.

I have retained the use of coupling for the main spinwave calculation, this makes a nice distinction between the functional roles.

*Exchange*: Real space component, user facing, single copy per interaction
*Coupling*: Fourier space component, used internally in the spinwave calculations, these come in pairs

**factional -> lattice_units**

I've renamed the functions in UnitCell that do the conversions to `lattice_units` as this is probably clearer

**TrivialSupercell -> TiledSupercell**

Just a class name change

**moment -> spin**

In most of the code it has been changed, but with two exceptions:
1) In the symmetry stuff, where it is actually moments that things operate on (though we can apply to spin)
2) In file names for icons - not worth changing


pyspinw/site.py
```diff
@@ -8,19 +8,19 @@
...
+    _unique_id_counter += 1
+    return _unique_id_counter
 
 class LatticeSite(SPWSerialisable):
     """A spin site within a lattice
 
     :param: i,j,k - *required* Fractional coordinates within unit cell
-    :param: mi,mj,mk - Magnetic moment along unit cell aligned axis
-    :param: supercell_moments - Magnetic moment for each propagation vector
+    :param: mi,mj,mk - Spin
```
mducle: I think that these should be `mx, my, mz - Spin components in the Cartesian directions defined by x||a, z perpendicular to a-b`, right?

mducle (changes requested):
Just one check - I think that in `LatticeSites` the atomic coordinates are in lattice units (which we agreed to call `i, j ,k`) but the spins are in Cartesian (which I think we agreed to use `x, y, z`). This is because the spin components in the `LatticeSites` are passed directly by `Hamiltonian` to `spinwave_calculation` without modification and `spinwave_calculation` requires the spin components in Cartesian to calculate the projections `z` and `eta` (which we call "rotations" in the code).

So could you change again `si, sj, sk` in `Sites` to `sx, sy, sz` and in particular update the docstring? It might also be good to add in the docstring that if users want to specify the spin components in lattice units that they use the `interface.generate_*_structure` functions?

Otherwise it looks good - and thanks for updating `__init__.py` with the new interface functions!

pyspinw/site.py
lucas-wilkins: I'll sort it out

lucas-wilkins: Done

mducle (approved):

PR #145, opened 2026-04-10
Non-hermitian spin wave calculation

mducle:
Implements the non-hermitian calculation (actually used when the spin hamiltonian is not positive semi-definite as we define the Hamiltonian to be Hermitian).

Unlike in the Matlab version, in Python/Rust we automatically switch between the two methods for computing the spin wave spectrum. The "hermitian" routine is used by default except when the Cholesky / LDL decomposition fails or detects that there are negative elements (below a threshold, currently set at `1e-7`) in the diagonal `D` matrix indicating that the Hamiltonian matrix is not positive semi-definite. If this happens the code switches automatically to using the "non-hermitian" algorithm.

There is now two tests of the non-hermitian routines using a square lattice antiferromagnet with spins in the $`a-b`$ plane but a significant axial anisotropy along $`c`$: one checking the consistency between Python and Rust calculations and one checking agreement between the Python calculation and the original Matlab code.


src/spinwave.rs
```diff
@@ -12,6 +12,8 @@ use crate::{Coupling, MagneticField, C64};
 
 /// Minimum energy that isn't just set for zero (in meV)
 const ZERO_ENERGY_TOL: f64 = 1e-12;
+const SINGULAR_FTOL: f64 = 1e-7;
+const SINGULAR_CTOL: C64 = C64::new(1e-7, 0.);
```
alexhroom, line 16: i think these should have docstrings (comments starting `///`) as it's not immediately clear what FTOL or CTOL mean

if we want them to be the same tolerance as each other just different types, I wonder if it might also be nice to have `SINGULAR_CTOL` be `C64::new(SINGULAR_FTOL, 0.)`?

src/spinwave.rs
```diff
@@ -407,6 +352,142 @@ pub fn calc_spinwave(
...
+        let d = ldl.B_diag().column_vector();
+        let s = ldl.B_subdiag().column_vector();
+        // The Bunch-Kaufmann algorithm works for indefinite matrices.
+        // But if the hamiltonian constructed above is not positive-semi-definite
+        // then g*ham will have imaginary eigenvalues and we need to use a different
+        // algorithm. We can check for semi-positive-definiteness here by checking if
+        // any off-diagonal element of the block-diagonal matrix are not zero
+        // or if any diagonal element is negative. I've not found any proof that the
+        // diagonal blocks are equivalent to eigenvalues but this post:
+        // https://mathoverflow.net/questions/84420 suggest it could be.
```
alexhroom, line 386: [Bunch & Kaufmann (1977)](https://www.jstor.org/stable/2005787) says on page 4
> a 2 x 2 block in D corresponds to a positive-negative pair of eigenvalues. This means that if A is positive definite then D will be diagonal.

and on page 14
> Suppose A = LDL^T, where L is unit lower triangular and D is diagonal. We shall show below that [the number of positive, negative, and zero eigenvalues] are equal to the number of positive, negative, and zero elements, respectively, on the diagonal of D.

src/spinwave.rs
```diff
@@ -109,39 +122,74 @@ fn calc_q_independent(
...
-/// - `spin_coefficients`: The spin coefficients matrix.
 /// - `couplings`: Slice of couplings between sites.
-/// - `Az`: Optional Zeeman term for the `A` matrix if a magnetic field is provided.
 ///
 /// # Returns
-/// The "square root" of the Hamiltonian matrix (K such that K K* = h(q)).
+/// The spin Hamiltonian matrix h(q)
 #[inline(always)]
-fn calc_sqrt_hamiltonian(
+fn calc_spin_hamiltonian(
```
alexhroom, line 182: this is nitpicky but while reading this code there were several times that I got confused because I'd skim-read `calc_spin_hamiltonian` as `calc_sqrt_hamiltonian` or vice versa. could this just be `calc_hamiltonian` or something more distinct?

src/spinwave.rs
```diff
@@ -407,6 +352,142 @@ pub fn calc_spinwave(
...
+    botleft *= -1.;
+    let mut topright = T.submatrix_mut(0, n_sites, n_sites, n_sites);
+    topright *= -1.;
+    let mut M = (T.adjoint() * eigvecs)
+        .diagonal()
+        .column_vector()
+        .to_owned();
+    M.iter_mut()
+        .for_each(|m| *m = (C64::ONE / (*m + SINGULAR_CTOL)).sqrt());
+    T = eigvecs * M.as_diagonal();
```
alexhroom, line 487: would probably be quite a small performance improvement but instead of copying `M` and performing a matrix-matrix multiplication, this could iterate over the columns of `eigvecs` from the reference of M - if you make `eigvecs` mutable, something like (**note this isn't code I've tested, just Rust-style pseudocode**)
```rust
let M = (T.adjoint() * eigvecs)
    .diagonal()
    .column_vector()
T = eigvecs
    .col_iter_mut()
    .zip(M)
    .map(|col, m| *col = col * (C64::ONE / m + SINGULAR_CTOL)).sqrt());
```

alexhroom (commented):
just some comments on the Rust as requested :) pretty solid functionally, I only have a small performance improvement and some readability/maintainability comments

PR #163, opened 2026-04-15
Changing OpenGL version from 3.3 to 4.1 for compatibility on macs

burniebarren:


burniebarren:
## Problem
On macOS, calling \`demos()\` crashes with:
\`\`\`
AttributeError: 'CrystalViewerWidget' object has no attribute 'axes_shader'
\`\`\`

This does not occur on Windows or Linux.

## Root Cause
macOS requires an OpenGL **core profile** context to use modern GL functions like \`glGenVertexArrays\`. Qt's default context creation didn't explicitly request a core profile, causing a fallback where these functions are unavailable.

## Solution
Applied 5 key fixes:
1. Set global OpenGL 4.1 core profile format before QApplication creation
2. Add widget-level format request in CrystalViewerWidget  
3. Implement graceful fallback initialization pattern
4. Replace deprecated glPushAttrib/glPopAttrib calls
5. Update GLSL shaders to version 410 core

## Testing
Verified on macOS - demos() opens viewer window without errors.

## Notes
It is indeed opening the viewers on mac, one needs to toggle off the magnet button on GUI for better experience. It may not be the optimal and best practice solution, I hope it nevertheless inspires for better fixes.

pyspinw/gui/viewer.py
```diff
@@ -1,10 +1,28 @@
 """ Viewer window """
 import ctypes
 import sys
+import os
+
+# Force desktop OpenGL before any Qt import (critical on macOS)
+os.environ.setdefault("QT_OPENGL", "desktop")
+os.environ.setdefault("QT_MAC_WANTS_LAYER", "1")
```
mducle, line 21: Could you move these three lines below all the `import` statements to fix the linting error, please?

mducle (changes requested):
@burniebarren thanks for this - it looks ok but could you move the `os.environ.setdefault` line below the imports to fix the [linting error](https://github.com/SpinW/pySpinW/actions/runs/24471627900/job/71563339666) and then rerun the linting check, please?

```
ruff check pyspinw
```

(you might have to install `ruff` with `pip install ruff` first).

pyspinw/gui/viewer.py
burniebarren: Modified and checked ruff on these files. Seems all clean.

burniebarren:
I see that the error `Resource not accessible by integration` means that the GitHub Action is trying to leave a comment on the Pull Request (to show the code coverage report), but it doesn't have the "write" permission required to post comments. The owner of the project needs to go to Settings > Actions > General and change the permissions to "Read and write permissions", or explicitly add this to the YAML file:
`permissions:
  pull-requests: write`

Let me know if I can do something further ✨.

mducle:
Thanks - it all looks ok to me but I think Lucas wants some changes... unfortunately his 2FA app is doing something weird and he can't log in to github.

mducle (approved):

AlbertRyu:
Hi all — I tried out @burniebarren's fix, and the only obvious remaining issue to me was High DPI scaling. I added a few lines so the resolution now adapts to the device pixel ratio on specific machines. The modification has been committed into this PR.

burniebarren:
Hey! I just realized that I merged the branch with the main one (or have I? I am not sure) 😅. I am not sure if the move is valid. Feel free to revert.

lucas-wilkins:
Thanks for doing this, it's great that you're getting involved!

There's a few things I'd like changed before this is merged, unfortunately
one of them is rather substantial.

* The guard statement that relies on hasattr should be replaced with a flag
for whether it is initialised (set to False in __init__ and True at the end
of initialiseGL), and initialiaseGL should not be called from within
painGL, just wait until it is ready.
* The try section in initialiseGL hides errors that should not be hidden.
logger.exception was the correct behaviour: we want errors in this part of
the program to cause it to fail.
* The big one: now things are in 4.1, there is a problem with
glLineWidth(4), this was "fixed" by setting it to 1.0. Unfortunately, this
makes the lines really small. There are two possible solutions, both are
not so simple. (option 1) start to draw wireframes in a more modern way. To
do it we need an "edge expanding" geometry shader that turns edge
information into a quads - often wireframes are drawn using a cutoff in
barycentric coordinates provided to a fragment shader, but this will not
work in this case, because we need the edges to expand beyond the sides of
the object drawn in front of it. (option 2) Display selections in a
different way, which means updating all the shaders *and* their interfaces.

I personally prefer option 1.

burniebarren:
👍🏽 I will check them out in the earliest convenience

lucas-wilkins:
These changes have been incorporated into #190

lucas-wilkins:
Closing as ported to #190

PR #184, opened 2026-05-06
Metadata for sites and exchanges

lucas-wilkins:
Adds metadata fields to sites and exchanges, allows for specifying colours and display sizes, and can infer them from the name.

Issue #173, opened 2026-04-23
'direct' analogue for generating magnetic supercell


Fellow-baka:
Hello, many thanks for you work. I started tinkering with pySpinW and can't find approach for 'direct' mode (like in MATLAB, directly define propagation vector and spins on each site in extended cell for arbitrary propagation vector) of generating extended magnetic structure for non zero propagation vector. (for some reasons I can't use helical one)
I also want to keep and define single unit cell with standard approach.

Here is some short example from spinW in MATLAB:

``` matlab
tst = spinw;
tst.genlattice('lat_const',[5 6 7],'angled',[90 90 90],'sym',58)
tst.addatom('label','Co2+','r',[0.0 0.0 0.0]);
tst.addatom('label','Co2+','r',[0.0 0.5 0.3]);

tst.gencoupling('maxDistance',3)

tst.addmatrix('label','J1','value',1,'color','red')
tst.addcoupling('mat','J1','bond',1)
tst.genmagstr('mode','direct','k',[0 1/2 1/2],'nExt', [1 2 2], 'S',[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0; 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0; 1 1 1 1 1 -1 -1 -1 -1 -1 -1 1 -1 -1 -1 -1 -1 1 1 1 1 1 1 -1 ]);

plot(tst,'range',[1 2 2],'tooltip',true)
```

For pyspinw I approached it in similar matter but was not able to generate supercell with arbitrary spins on multiple sites in supercell.

``` python
from pyspinw import *

unit_cell = UnitCell(5,6, 7, 90, 90, 90)
sg = spacegroup("P n n m")

s1 = LatticeSite(0, 0, 0, name="Co_1")
s2 = LatticeSite(0, 0.5, 0.3, name="Co_2")

sites = [s1, s2]

s = generate_structure(unit_cell,
                               positions=sites,
                               spins=[[0,0,1],[0,0,-1]])

view(s)
```

lucas-wilkins:
If I understand correctly, what you are asking for is one or more propagation vectors each associated with only one site.

The way I would approach this is using the summation supercell along the following lines, assuming you want two completely different helices in the xy plane:

```
[...]

PR #185, opened 2026-05-06
CIF file reader

### LINKED ISSUE #174: CIF and mCIF
The original implementation I did of CIF stuff was incomplete and now in the trash.

We do need this, although it would probably be better to do this before/at the same time as adding more metadata to sites (e.g. colours, atom types, etc).

lucas-wilkins:
Adds a CIF file reader (currently not mCIF)


lucas-wilkins:
Extends #184 and #173

pyspinw/cif.py
```diff
@@ -0,0 +1,95 @@
...
+    try:
+        name = names[entry_index]
+    except IndexError:
+        raise ValueError(f"Entry index out of range (file has {len(names)} entries)")
+
+    data = cif[name]
+
+    # Get the spacegroup and lattice parameters
+
+    spacegroup_name = data["_symmetry_space_group_name_H-M"]
```
mducle: So this key could be a lot of things... and moreover the [entry for the key](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Isymmetry_space_group_name_H-M.html) itself says that it has been superseded by [_space_group_name_H-M_alt](https://www.iucr.org/__data/iucr/cifdic_html/1/cif_core.dic/Ispace_group_name_H-M_alt.html) - we should be checking both these keys (e.g. I tried it with a VESTA generated CIF from [an old training course](https://github.com/mducle/edatw24/blob/main/spinw_files/BiFeO3.cif) and it failed because VESTA uses the new key).

The equivalent [ASE code](https://gitlab.com/ase/ase/-/blob/master/ase/io/cif.py?ref_type=heads#L325-329) checks those two keys as well as [_space_group.Patterson_name_H-M](https://www.iucr.org/__data/iucr/cifdic_html/2/cif_sym.dic/Ispace_group.Patterson_name_H-M.html).

We might also want to check [_space_group.IT_number](https://www.iucr.org/__data/iucr/cifdic_html/2/cif_sym.dic/Ispace_group.IT_number.html)...

Also, since we have a dependency on ASE maybe we can just use the ASE CIF parser?

mducle (changes requested):
CIF grammar is horrendous and changeable and we need to check more than one key for the spacegroup...

pyspinw/cif.py
lucas-wilkins: Done

mducle (approved):

PR #189, opened 2026-05-18
188 return eigenvectors from calculations

### LINKED ISSUE #188: Return eigenvectors from calculations
We need to be able to get the eigenvectors from the calculations. This is needed for:

* Multimagnon stuff, e.g. #170 
* (assuming my strategy works) assigning modes correctly #186 #171

lucas-wilkins:
Exposes the transformation matrices and cleans up some of the interfaces between main calculation and the user side


mducle (approved):

lucas-wilkins:
TODO: fix the issue with `_energies_and_intensities` vs `energies_and_intensities`

PR #190, opened 2026-05-21
114 viewer todos and proposed features

### LINKED ISSUE #114: Viewer TODOs and proposed features
The viewer still has a number of planned features that are still TODO:

- [x] Add atom only view
- [x] Colour by atom type
- [ ] DM and complex interaction representations
- [ ] Anisotropies
- [ ] Show hide items
  - [ ] Individual items
  - [ ] Slices
- [x] The scaling sliders should be relative to the unit cell size, not absolute
  - [x] Also a bug where the atoms/spins get bigger each time it runs
- [x] Allow a non-blocking call and multiple instances
- [x] Just generate an image
  - [x] Save image functionality
- [x] Change background color
- [x] Change default colors
- [x] Change selection colors
- [x] Fix the macOS rendering issues
- [ ] Scale structures on input
  - [ ] Deal with incommensurate structures
- [x] Orthogonal / Perspective toggle
Add any other features here

lucas-wilkins:
Some of the proposed viewer changes have been implemented:

- [x] Add atom only view
- [x] Colour by atom type
- [x] The scaling sliders should be relative to the unit cell size, not absolute
  - [x] Also a bug where the atoms/spins get bigger each time it runs
- [x] Allow a non-blocking call and multiple instances
- [x] Just generate an image
  - [x] Save image functionality
- [x] Change background color
- [x] Change default colors
- [x] Change selection colors
- [x] Fix the macOS rendering issues

Other's remain on #114 - but I think this will do for this PR


lucas-wilkins:
@AlbertRyu Could you check the high DPI changes work on mac? (that things are the right scale and axes in the right position)

AlbertRyu:
@lucas-wilkins I attempted to launch the viewer on branch 114. There's an error message of missing SVG file. 

The error message is as following:

qt.svg: Cannot open file '/Users/yunxiaoliu/SyncDokumente/Projects/pySpinW/pyspinw/gui/icons/svg/camera.svg', because: No such file or directory

```text
Exception                                 Traceback (most recent call last)
Cell In[5], [line 13](vscode-notebook-cell:?execution_count=5&line=13)
      9 # Create a magnetic structure (this could include a spacegroup, or
     10 # supercell structure, but we don't do so here)
     11 structure = Structure([x], unit_cell=unit_cell)
     12 
---> [13](vscode-notebook-cell:?execution_count=5&line=13) view(structure)
     14 
     15 # A single exchange between atoms in neighbouring unit cells (in the x direction)
     16 exchanges = [HeisenbergExchange(x, x, cell_offset=(1,0,0), j=-1)]

File ~/SyncDokumente/Projects/pySpinW/pyspinw/gui/viewer.py:306, in show_object(object, block)
    300 app.setWindowIcon(png_icon("pyspinw"))
    302 # Useful for checking particular display options
    303 # app.styleHints().setColorScheme(Qt.ColorScheme.Dark)
    304 # app.styleHints().setColorScheme(Qt.ColorScheme.Light)
--> [306](https://file+.vscode-resource.vscode-cdn.net/Users/yunxiaoliu/SyncDokumente/Projects/playground/~/SyncDokumente/Projects/pySpinW/pyspinw/gui/viewer.py:306) viewer = Viewer(object)
    307 viewer.setWindowTitle("Hamiltonian Viewer")
    308 viewer.resize(800, 600)

File ~/SyncDokumente/Projects/pySpinW/pyspinw/gui/viewer.py:82, in Viewer.__init__(self, hamiltonian, initial_rotation, initial_distance, display_options, parent)
     78 render_model = RenderModel(hamiltonian)
     80 layout = QVBoxLayout()
---> [82](https://file+.vscode-resource.vscode-cdn.net/Users/yunxiaoliu/SyncDokumente/Projects/playground/~/SyncDokumente/Projects/pySpinW/pyspinw/gui/viewer.py:82) self.toolbar = DisplayOptionsToolbar(initial_display_options=display_options)
...
---> [17](https://file+.vscode-resource.vscode-cdn.net/Users/yunxiaoliu/SyncDokumente/Projects/playground/~/SyncDokumente/Projects/pySpinW/pyspinw/gui/icons/iconload.py:17)     raise Exception(f"SVG is invalid: {svg_path}")
     20 # Render SVG to pixmap
     21 pixmap = QPixmap(size, size)

Exception: SVG is invalid: /Users/yunxiaoliu/SyncDokumente/Projects/pySpinW/pyspinw/gui/icons/svg/camera.svg`
```

lucas-wilkins:
> @lucas-wilkins I attempted to launch the viewer on branch 114. There's an error message of missing SVG file.
> 
> The error message is as following:
> 
> qt.svg: Cannot open file '/Users/yunxiaoliu/SyncDokumente/Projects/pySpinW/pyspinw/gui/icons/svg/camera.svg', because: No such file or directory


Woops, forgot to commit the icons. Done now.

AlbertRyu:
@lucas-wilkins The viewer is now rendering nicely at high DPI. There is just one remaining glitch.

I recorded a short demo showing the behavior of all the buttons, including the issue mentioned above. When I click the `show lattice aligned axes` button, it continuously raises the error message shown in the demo. Shortly afterward, my local Python process crashes, and I have to force-kill it.

```text
Error calling Python override of QOpenGLWidget::paintGL(): 
---------------------------------------------------------------------------
ArgumentError                             Traceback (most recent call last)
File ~/SyncDokumente/Projects/pySpinW/pyspinw/gui/crystalview.py:393, in CrystalViewerWidget.paintGL(self)
    390 glDisable(GL_DEPTH_TEST)
    392 # Lattice axes viewport
--> [393](https://file+.vscode-resource.vscode-cdn.net/Users/yunxiaoliu/SyncDokumente/Projects/playground/~/SyncDokumente/Projects/pySpinW/pyspinw/gui/crystalview.py:393) glViewport(
    394     (self.width() - self.axes_size - self.axes_padding) * dpr,
    395     self.axes_padding * dpr,
    396     self.axes_size * dpr,
    397     self.axes_size * dpr
    398 )
    400 self.axes_shader.camera = self.camera
    402 if self.display_options.orthogonal_lattice_axes:

ArgumentError: argument 1: TypeError: 'float' object cannot be interpreted as an integer
```
https://github.com/user-attachments/assets/aa495554-fdeb-46b3-9046-19370f7e11c2

lucas-wilkins:
I've fixed that bug.

It doesn't look like the hovering is working correctly. I'll see if I've done something dumb.

It should look like this:

https://github.com/user-attachments/assets/92f6e95f-7ad4-44be-8aaa-bec3e4f11eef

Uploading Recording 2026-06-02 153243.mp4…

lucas-wilkins:
@AlbertRyu Does my latest change fix it?

AlbertRyu:
@lucas-wilkins 👍 The last change fixed it. And the hovering also looks correct. A short video is attached to show the behavior. 

There's a harmless warning I forgot to mention in the previous message. 
```test
qt.qpa.fonts: Populating font family aliases took 477 ms. Replace uses of missing font family "Sans-serif" with one that exists to avoid this cost.
```

I also noticed that, when I launch the viewer from a Jupyter notebook, the kernel still crashes every time I close the viewer(without any error, just crash), requiring me to force-quit the process. However, when I launch the viewer from a .py script, it exits gracefully without any problems.

I'm not sure whether this is specific to my local environment. If this behavior is reproducible across different setups, it may cause some confusion and inconvenience for users who work from Jupyter notebooks.



https://github.com/user-attachments/assets/f55774c4-5686-4a33-b547-fcae4907aa34

lucas-wilkins:
> There's a harmless warning I forgot to mention in the previous message.

Might be a lot of effort to make this go away, don't think its a priority, I've created an issue for it - #192 .

> I also noticed that, when I launch the viewer from a Jupyter notebook, the kernel still crashes every time I close the viewer(without any error, just crash), requiring me to force-quit the process. However, when I launch the viewer from a .py script, it exits gracefully without any problems.

Hmmmm, I wonder if its actually crashing or just not handing back control to jupyter?

AlbertRyu:
> if its actually crashing or just not handing back control to jupyter?

@lucas-wilkins I investigated this a bit further and I believe this is primarily a macOS-specific issue. I'm not sure whether we can address it on our side; if we can, it would probably belong in a separate issue.

What appears to happen is:
1. When a Jupyter kernel starts, it runs as a background Python process.
2. When the viewer is launched, the same process (same PID) creates a QApplication and is registered by macOS as a GUI application, which causes it to appear in the Dock.
3. When the viewer window is closed, the GUI is torn down, but the underlying process remains alive because it is still the Jupyter kernel serving notebook requests.
4. This seems to leave macOS in a confused state: the process is still registered as a GUI application, but it is no longer behaving like an active GUI app. As a result, macOS marks it as unresponsive and periodically attempts to interact with it as a GUI application.

This has two side effects:
1. The viewer cannot be opened a second time without restarting the kernel, because the same PID is already associated with the previous GUI application state.
2. If the kernel is left running, macOS periodically tries to communicate with the stale GUI application, which can cause temporary system hangs (spinning beach ball). The only workaround I have found is to force-quit the process, which also terminates the Jupyter kernel.

lucas-wilkins:
> I believe this is primarily a macOS-specific issue.

Yeah, I can't reproduce this on my machine

I've added some more functionality for the viewer, you should be able to run with `block=False` where it should just keep things running in the background instead of taking over the thread entirely, that might be a workaround?

Related thing, does `snapshot` work?

lucas-wilkins:
Think this will do for now

mducle:
>     1. The viewer cannot be opened a second time without restarting the kernel, because the same PID is already associated with the previous GUI application state.
> 
>     2. If the kernel is left running, macOS periodically tries to communicate with the stale GUI application, which can cause temporary system hangs (spinning beach ball). The only workaround I have found is to force-quit the process, which also terminates the Jupyter kernel.

We could maybe get around this by changing the code in `show_hamiltonian` to only create the app once and assign it to a global variable. Then subsequent calls to `show_hamiltonian` can just redraw the main window and call `show()` on it.

pyspinw/gui/viewer.py
```diff
@@ -1,34 +1,94 @@
...
+
         render_model = RenderModel(hamiltonian)
 
         layout = QVBoxLayout()
 
-        self.toolbar = DisplayOptionsToolbar()
+        self.toolbar = DisplayOptionsToolbar(
+            initial_display_options=display_options,
+            save_config_on_exit=self.save_config_on_exit,
+            max_size_object_scaling=hamiltonian.structure.unit_cell.main_diagonal_length)
```
mducle: I think this is a bit large - basically at the maximum size each object is the size of the cell, which I'm not sure anyone really wants. It also means that it is a bit fiddly to control at the small end. I think maybe half this size would be better for the maximum.

We could possibly make this scaling an option which users can set.

mducle (approved):
Just one aesthetic comment - I found that I only used about 1/4 of the range of the spin/exchange slider because the maximum (basically the unit-cell size) is too big. So it's a bit fiddly to get the right size.

Otherwise everything looks good.

I checked on Linux with Jupyter and found that I could plot multiple structures in different cells with the same kernel, so the global `_VIEWER` reference works... so I don't know what the additional problem with MacOS is...

pyspinw/gui/viewer.py
lucas-wilkins: Yeah, I get what you mean, however,

If there is a single atom in the unit cell, then this is about right.

I thought about limiting it by the volume / number of atoms, but then, atoms can have very different sizes.
I did mess around with some non-linear scaling, which didn't work out too well, it could be revisited.

> We could possibly make this scaling an option which users can set.

A setting for the settings ;)

lucas-wilkins: Saying that, probably main diagonal is too big even for the single atom case. Really, it can't be bigger than the smallest of the side lengths, which for a cubic cell, would be reduction by a factor of 1.73, and even more for irregular cells.

lucas-wilkins: > I checked on Linux with Jupyter and found that I could plot multiple structures in different cells with the same kernel, so the global _VIEWER reference works... so I don't know what the additional problem with MacOS is...

Yeah, I've sorted out access to the RSE mac, so I can have a look on that, probably.

mducle: Cool thanks... using the smallest lattice parameter is good enough for me - I was checking it with the kagome example with a hexagonal cell with a large `c` which makes it even worst.

mducle (approved):

PR #196, opened 2026-06-07
Symmetry analysis

### LINKED ISSUE #193: Symmetry Analysis
Feature parity with MATLAB spinW for symmetry checking and building systems

Current plan:
 - [ ] Add a function `allowed_exchanges(site, site)`
 - [ ] Add a function to return all symmetries that transform one pair of points into another
 - [ ] Add `.verify_symmetry(spacegroup)` to exchanges
 - [ ] Add `.verify_symmetry()` to hamiltonian
 - [ ] Add `.symmetry_updated(...)` to exchanges
 - [ ] Add bond based coordinates to DM interaction

### LINKED ISSUE #212: Feature: Checking for Spacegroup + UnitCell consistency
A method to check that unit cell is consistent with the spacegroup.

Much of what is needed for this already exists, however, the database of unit cell constraints only applies to the 230 standard settings, not the full 500 odd hall numbers. Since we added support for alternate settings, and since these are represented as unique spacegroups and not mapped to the standard setting, the unit cell constraints need extending in some way to account for different permutations of axes.

lucas-wilkins:
Adds symmetry checking, information, and ability to use symmetry to create exchanges from an existing one by symmetry.
It also adds some functions to make selecting sites easier

Adds methods to spacegroups, space operations, sites, structure, exchange, anisotropy, and hamiltonian. These can check symmetry, give information about constraints on matrices, automatically fill, do transforms etc.

The following is still outstanding, and actually a bit of a problem, will make a new issue for this: "Finding nearest neighbours is currently very inefficient. This needs to be made better"

During this I discovered a bug in the CIF file reader, this includes a fix for that too.

Includes, and closes #212


lucas-wilkins:
It wasn't the plan, but making tests for this has involved most of the work for #212, adding that in.

debug/exchange_copying.py
```diff
@@ -0,0 +1,19 @@
...
+cell = UnitCell(1,1,1)
+
+struct = Structure(sites=[centre, offset], unit_cell=cell, spacegroup=p4mmm)
+
+struct.print_summary()
+
+other_offset = struct.site_by_name("offset [1]")
+exchange_1 = DMExchange(centre, offset, d_x=1, d_y=1, d_z=1)
+
+exchange_2 = exchange_1.symmetry_copy(centre, other_offset, p4mmm)
```
mducle: Is `symmetry_copy` still a WIP? This script gives a couple of `AttributeError`s and even when I fixed that it gives a 

```
ValueError: Exchange does not obey symmetry constraints, cannot use symmetry to copy
```

Also I noticed that there are no tests for `symmetry_copy`...

----

w.r.t. the `AttributeError`, the function signature seems to have changed since you wrote this script - it expects a `Structure` as the first argument and doesn't want a `Spacegroup` as the last argument - so I changed it to:

```
exchange_2 = exchange_1.symmetry_copy(struct, centre, other_offset)
```

And then also had to change the code in [exchange.py#L341](https://github.com/SpinW/pySpinW/blob/193-symmetry-analysis/pyspinw/exchange.py#L341) to `if self.obeys_symmetry(structure):` since `obeys_symmetry` expects a `Structure` rather than a `Spacegroup`.

debug/hamiltonian_symmetry_check_debug.py
```diff
@@ -0,0 +1,17 @@
...
+
+cell = UnitCell(1,1,1)
+
+structure = Structure([x,y], cell, sg)
+
+hamiltonian = Hamiltonian(structure, [HeisenbergExchange(x, y, j=1, cell_offset=(0,0,1))])
+
+hamiltonian.print_summary()
+
+hamiltonian.complete_symmetry()
```
mducle, line 17: This method `complete_symmetry` doesn't exist in the current code - is it meant to? If not maybe remove this file?

pyspinw/anisotropy.py
```diff
@@ -69,6 +72,130 @@ def anisotropy_matrix(self) -> np.ndarray:
...
+        # There might be more than one symmetry operation that maps the pair of sites
+        #  however, the effect on the anisotropy should be the same for all these operations,
+        #  this means we can just pick an arbitrary one.
+        # If this turns out not to be the case, then anisotropy itself does not need to obey the
+        # symmetry constraints
+
+        spacegroup = structure.spacegroup
+        unit_cell = structure.unit_cell
+
+        if self.obeys_symmetry(spacegroup):
```
mducle: This should be `if self.obeys_symmetry(structure)` - I guess the syntax of `obeys_symmetry` changed during development? Also there's no test for this method (`symmetry_copy`).

pyspinw/exchange.py
```diff
@@ -265,6 +259,247 @@ def is_symmetric(self):
...
+        # There might be more than one symmetry operation that maps the pair of sites
+        #  however, the effect on the exchange should be the same for all these operations,
+        #  this means we can just pick an arbitrary one.
+        # If this turns out not to be the case, then exchange itself does not need to obey the
+        # symmetry constraints
+
+        spacegroup = structure.spacegroup
+        unit_cell = structure.unit_cell
+
+        if self.obeys_symmetry(spacegroup):
```
mducle: Same comment as above for `Anisotropy.symmetry_copy`.

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+$a s_1 + b s_2 = a g(s_1) + b g(s_2)$
+
+Then applying the definition of the transformation
+
+$= a M s_1 + a T + b M s_2 + b T = M (a s_1 + b s_2) + (a+b) T$
+
+and as long as $a + b = 1$
+
+$ = g(a s_1 + b s_2)$
```
mducle: Need to remove the first space here for it to render `$= g(a s_1 + b s_2)$`

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+$= (k-1) M s_1 + k M s_2 + T = g((k-1) s_1 + k s_2)$
+
+and in this case we can see that there is a further restriction in this case, so that
+
+$g(k s_1 + (1-k) s_2) = g((k-1) s_1 + k s_2)$
+
+i.e. $k = 1-k$, or $k = 1/2$. That is to say, the symmetries when swapped are symmetries of the midpoint.
+
+This doesn't prove that all symmetries of the midpoint are symmetries of the point pair, just that symmetries
+of the point pair are symmetries of the midpoint.
```
mducle, line 91: Maybe italicise this point? Which suggests that the methods used in Matlab SpinW could have placed more restrictions when calculating symmetry allowed exchange terms than there was.

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+A numerical solution can be obtained by Gaussian elimination / row-reduction.
+Essentially, we need to "flatten the equation"
+the unswapped constraints are calculated using the following vectorisation identity
+
+$\text{vec}(AXB)=(B^T \otimes A)\text{vec}(X)$
+
+where $\text{vec}$ denotes the vectorisation, i.e. such that
+
+$\text{vec} \left(\begin{array}{ccc} a & b & c \\ d & e & f \\ g & h & i \end{array}\right) = \left(\begin{array}{c} a \\ b \\ c \\ d \\ e \\ f \\ g \\ h \\ i  \end{array}\right)$
```
mducle: This equation is not rendering, but I could get it to render by changing the first dollar to `` $` `` and the last to `` `$ `` (add backticks after the first and before the last)

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+The swapped constraints need a "commutation matrix", $K$, defined such that
+
+$\text{vec}(P^T) = K \text{vec}(P)$
+
+With this, for the swapped case we get
+
+$((M \otimes M) K - I) (R \otimes R) V = 0$
+
+To get a reduced system of equations for out matrix entries we build a $9$-by-$9n$
```
mducle: Change to ``$9$-by-$`9n`$`` to render (no idea why original doesn't).

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+### In terms of symmetric and antisymmetric components
+
+This will be enough to give us a simplified set of equations for the system, 
+but it is not the preferred form for users. It's more understandable to consider 
+the symmetric and antisymmetic parts of the matrix separately.
+
+Let's now write $J$ in terms of matrix entries, with the aim of making a simple linear system in the form $Y = MX$
+
+$J = \left( \begin{array}{ccc} a & b & c \\ b & d & e \\ c & e & f \end{array} \right) + \left( \begin{array}{ccc} 0 & z & -y \\ -z & 0 & x \\ y & -x & 0 \end{array} \right)$
```
mducle: Again need backticks after the first dollar and before the last dollar here.

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+We then want an equation to relate the entries to $V$ (dots used instead of zeros for readability)
+
+$V = \left(\begin{array}{ccccccccc}
```
mducle: And again backticks here.

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+\cdot & \cdot & 1 & \cdot & \cdot & \cdot & \cdot & 1 & \cdot \\
+\cdot & \cdot & \cdot & \cdot & 1 & \cdot & -1 & \cdot & \cdot \\
+\cdot & \cdot & \cdot & \cdot & \cdot & 1 & \cdot & \cdot & \cdot \\
+\end{array}\right) \left(\begin{array}{c} a \\ b \\ c \\ d \\ e \\ f \\ x \\ y \\ z  \end{array}\right)$
+
+To get a system of equations in these terms, we can premultiply before doing the reduction.
+
+## Copying Exchanges
+
+### Uniqueness of symmetry based copying
```
mducle, line 199: I don't understand this section... could you be clearer? Are you saying here that $h$ and $g$ are the same? (That is, all the symmetry operations of a given space group which maps $(s_1, s_2)$ to $(t_1, t_2)$ are the same?) - and how does this relate to copying exchanges?

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
```diff
@@ -0,0 +1,287 @@
...
+
+which we can vectorise to give
+$(M \otimes M - I) \;\text{vec}A = 0$
+
+where $\text{vec}A$ is the vectorised version of $A$.
+
+But we only have 6 degrees of freedom, to get this we can use a truncated version of the matrix $V$ above,
+which we can produce by truncating the matrix in $V$ above, i.e.
+
+$\text{vec}A = \left(\begin{array}{c} a \\ b \\ c \\ b \\ d \\ e \\ c \\ e \\ f  \end{array}\right) = \left(\begin{array}{ccccccccc}
```
mducle: Again backticks needed here.

pyspinw/symmetry/group.py
```diff
@@ -187,6 +180,152 @@ def implied_sites_for(self, site: LatticeSite) -> list[ImpliedLatticeSite]:
...
+                             site_2: LatticeSite | ArrayLike,
+                             unit_cell: UnitCell,
+                             do_print: bool = True) -> ExchangeMatrixConstraints:
+        """ Get the details of the allowed exchange matrices"""
+        if not isinstance(site_1, LatticeSite):
+            try:
+                site_1 = LatticeSite(i=float(site_1[0]),
+                                     j=float(site_1[1]),
+                                     k=float(site_1[2]),
+                                     name="tmp_site_1")
```
mducle, line 249: Should there be a `LatticeSite.coerce` method which does this?

pyspinw/symmetry/symmetry_checking.py
```diff
@@ -1,63 +1,378 @@
...
-            #  As the magnitude of S can be different when actually running the calculation,
-            #  this can only happen when the constant is zero,
-            #  which is when R1 M R2 is antisymmetric
-
-            pass
+from pyspinw.util import rotation_matrix
+
+
+def rref(A, tol=1e-12):
+    """ Reduced row echelon form - not in scipy or numpy"""
```
mducle, line 9: I thought this was quite surprising but looking into it a bit more it seems that's because LAPACK just has LU-decomposition but their formulation does not guarantee that the U matrix produced is in RREF. It doesn't matter if you want to solve a system of equation but here it does... 🤷

pyspinw/symmetry/symmetry_checking.py
```diff
@@ -1,63 +1,378 @@
...
+    def __init__(self, n: int):
+        self.free = None
+        self.zero = None
+        self.constraints = None
+        self.n = n
+
+    def _matrix_form_strings(self) -> tuple[list[str], list[str]]:
+        """ Gets the strings used to build matrices and constraint equations"""
+        strings = [" 0" for _ in range(self.n)]
+        characters = PopString("abcdefghi")
```
mducle: Could you just use:

```
characters = ['i', 'h', 'g', 'f', 'e', 'd', 'c', 'b', 'a']
```

Instead of creating the helper class `PopString`? `List` has a `pop()` method but it operates from the end rather than the start.

pyspinw/anisotropy.py
```diff
@@ -69,6 +72,130 @@ def anisotropy_matrix(self) -> np.ndarray:
...
+            if i >= len(new_anisotropies):
+                break
+
+            to_keep: Anisotropy = new_anisotropies[i]
+            to_remove: list[int] = []
+            for j, to_check in enumerate(new_anisotropies[i+1:]):
+
+                # One anisotropy per site
+                if to_keep.site == to_check.site:
+                    to_remove.append(j + i + 1)
```
mducle, line 185: Why is it just the site which is being checked and not the `anisotropy_matrix` values? (Could there ever be a case where the code duplicates the same site but gives a different value for `anisotropy_matrix`?)

pyspinw/anisotropy.py
```diff
@@ -69,6 +72,130 @@ def anisotropy_matrix(self) -> np.ndarray:
...
+            # Remove the ones we need to remove, as its ordered we can do
+            to_remove.reverse() # in place reverse
+            for j in to_remove:
+                del new_anisotropies[j]
+
+        # Remove the original input anisotropy if not included, it will always still be at the front
+        if not include_original:
+            new_anisotropies = new_anisotropies[1:]
+
+        return [specialise_anisotropy(anisotropy) for anisotropy in new_anisotropies]
```
mducle, line 197: The above code (L169-195) could be replaced by a `set(new_anisotropies)` if you define a `__hash__` and `__eq__` method for the `Anisotropy` class:

```py
    def __hash__(self):
        return hash((self.name, self.site.name))

    def __eq__(self, other):
        return self.site == other.site and (self.anisotropy_matrix == other.anisotropy_matrix).all()
```

pyspinw/cif.py
```diff
@@ -22,18 +22,27 @@ def parse_float_entry(s: str):
...
     return float(s)
 
-def load_cif(filename: str | pathlib.Path, supercell: Supercell = TiledSupercell(), entry_index: int=0, verbose=False):
+class LoadError(Exception):
+    """ Error thrown when loading fails"""
+
+def load_cif(filename: str | pathlib.Path,
+             supercell: Supercell = TiledSupercell(),
+             entry_index: int=0,
+             verbose=True) -> Structure:
```
mducle: Umm... could you change the default of `verbose` back to `False` so as to not confuse users?

pyspinw/anisotropy.py
```diff
@@ -69,6 +72,130 @@ def anisotropy_matrix(self) -> np.ndarray:
...
+                            anisotropy_matrix=new_anisotropy_matrix,
+                            #name=f"{self.name} [{op.text_form}]"
+                              )
+
+        else:
+            raise ValueError("Anisotropy does not obey symmetry constraints, cannot use symmetry to copy")
+
+    def symmetry_fill(self,
+                      structure: "Structure",
+                      include_original=False,
```
mducle, line 139: I'm not sure of the use case for this `include_original=False` - the only place where `symmetry_fill` is used in the code is in `Hamiltonian.symmetry_filled` which uses `include_original=True`, and I would imagine that users would want this kind of behaviour (get _all_ symmetry equivalent anisotropies) rather than the `include_original=False` behaviour... that seems more like what you'd expect from `symmetry_copy` and we kind of thought that this would be a niche use-case...?

pyspinw/exchange.py
```diff
@@ -265,6 +259,247 @@ def is_symmetric(self):
...
+                            exchange_matrix=new_exchange_matrix,
+                            name=f"{self.name} [{op.text_form}]",
+                            cell_offset=CellOffset.coerce(cell_offset))
+
+        else:
+            raise ValueError("Exchange does not obey symmetry constraints, cannot use symmetry to copy")
+
+    def symmetry_fill(self,
+                      structure: "Structure",
+                      include_original=False,
```
mducle, line 366: Same comment as for `Anisotropy.symmetry_fill` - I'm not sure if this is a useful default...

pyspinw/hamiltonian.py
```diff
@@ -25,15 +26,35 @@
...
+    """ Used to create defaultdicts of defaultdicts of sets """
+    return defaultdict(set)
+
+def _defaultsetdictdict():
+    """ Used to create defaultdicts of defaultdicts of defaultdict of sets """
+    return defaultdict(_defaultsetdict)
+
+def _defaultsetdictdictdict():
+    """ Used to create defaultdicts of defaultdicts of defaultdict of defaultdict of sets """
+    return defaultdict(_defaultsetdictdict)
```
mducle: These functions don't seem to be used anywhere in the code - in `_combine_exchanges` you just create a `defaultdict` directly - maybe these can be removed?

pyspinw/structure.py
```diff
@@ -280,21 +322,131 @@ def print_summary(self):
...
+        """
+        all_neighbors = self.all_neighbours(
+                site = site,
+                n=n,
+                element=element,
+                direction_filter=direction_filter)
+
+        return all_neighbors[-1]
+
+    def one_neighbour(self,
```
mducle, line 437: Not sure this function is useful - why not just let users compute all the neighbours since that's what we're doing here anyway? Also it's not used anywhere in the code and is not tested.

mducle (changes requested):
Thanks for this... and sorry the review took so long. I've got a couple of suggested changes... mostly minor.

I remember you mentioning that the neighbour computation is slow - [scikit-learn](https://scikit-learn.org/stable/modules/neighbors.html#nearest-neighbor-algorithms) has some tree-based algorithm for finding neighbours from point clouds but I not sure if we should have it as a dependency... or we can copy their [tree class](https://github.com/scipy/scipy/blob/v1.18.0/scipy/spatial/_kdtree.py#L276-L958) which is around 700 lines...

pyspinw/structure.py
lucas-wilkins: I use it, makes more sense when you are working with `symmetry_filled`.

pyspinw/hamiltonian.py
lucas-wilkins: probably, they're quite a thing!

pyspinw/anisotropy.py
lucas-wilkins: the main function I expect people to use is `symmetry_filled` on the hamiltonian, the use case for this is when you already have a defined exchange, in which case you've probably already assigned it to a variable. You probably want a list of the others, not the one you've assigned, in this use case.

pyspinw/exchange.py
lucas-wilkins: same response

pyspinw/cif.py
lucas-wilkins: yeah

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
lucas-wilkins: ugh, markdown latex

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
lucas-wilkins: There is actually a second constraint which I've not added to the docs, that I think might do the other half of "iff" proof (only considering transformations that don't have a translation).

docs/developers/notes/008_exchange_and_anisotropy_symmetries.md
lucas-wilkins: because of course

pyspinw/anisotropy.py
lucas-wilkins: I'll look into tests, maybe I'll get rid of these functions entirely, as I'm not sure how useful they are. Maybe a bit for some debugging.

debug/hamiltonian_symmetry_check_debug.py
lucas-wilkins: yeah, its been renamed to `symmetry_filled`

debug/exchange_copying.py
lucas-wilkins: yeah, mostly a defunct debug file.

pyspinw/hamiltonian.py
lucas-wilkins: Removed

pyspinw/exchange.py
lucas-wilkins: fixed

pyspinw/cif.py
lucas-wilkins: done

pyspinw/hamiltonian.py
lucas-wilkins: gone

lucas-wilkins: don't know why the resolve button doesn't work

pyspinw/symmetry/group.py
lucas-wilkins: Think its probably fine for now, maybe if it needs to happen a third time.

pyspinw/symmetry/symmetry_checking.py
lucas-wilkins: done

mducle (approved):
Looks good now. Thanks!

PR #198, opened 2026-06-16
Fixed bug with element assignment

### LINKED ISSUE #197: Handle ions in CIF files
Currently, entries of the form [M][n+] and [X][n-] in CIF files cause an error during the assignment of element.

lucas-wilkins:
Fixes bug where element assignment is confused by ions.

Problem was that the atom label was getting assigned to the element, after the bit that puts it in standard form.

### DIFF
--- a/pyspinw/sitemeta.py
+++ b/pyspinw/sitemeta.py
@@ -15,7 +15,7 @@ def _build_regex():
     symbols_sorted = sorted(chemical_symbols[1:], key=lambda x: (-len(x), x))
 
     # Build regex
-    pattern = r"^(?:{})".format("|".join(symbols_sorted))
+    pattern = r"^\s*({})(?:[^a-zA-Z]?.*)?$".format("|".join(symbols_sorted))
     return re.compile(pattern)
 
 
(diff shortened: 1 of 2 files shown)


pyspinw/sitemeta.py
```diff
@@ -15,7 +15,7 @@ def _build_regex():
     symbols_sorted = sorted(chemical_symbols[1:], key=lambda x: (-len(x), x))
 
     # Build regex
-    pattern = r"^(?:{})".format("|".join(symbols_sorted))
+    pattern = r"^\s*({})(?:[^a-zA-Z]?.*)?$".format("|".join(symbols_sorted))
```
mducle: I'm not quite sure why you have the pattern `[^a-zA-Z]?` which took me a couple of minutes in the docs to figure out but it seems to mean "match something which is not a lower or upper case letter zero or once" but then you have `.*` which is "match any character any number of times" - the `.*` seems to negate the `[^a-zA-Z]` to me... 

Or do you mean `[^a-zA-Z]+` which means "match something which is not a lower or upper case letter at least once" - as far as I can tell all site symbols seem to follow the pattern of `<chemical element><number>` so this would make more sense to me in that it would exclude things like `AuX` but include `Au11` (and spit out `Au` as the chemical element in both cases). But if we make the assumption that it's always a number in the site label, maybe we can just use something like `r"^\s*({})(?:[0-9]*)?$".format("|".join(symbols_sorted))`? (which would exclude weirdness `Au!`)

mducle (changes requested):
Just a query on the regex... maybe I'm not understanding it right?

pyspinw/sitemeta.py
lucas-wilkins: Done

mducle (approved):

PR #202, opened 2026-06-26
Two quick fix of import

AlbertRyu:
The following two minor changes are to remove warnings while auto-generating sphinx docs.

* Changed the import of `Instrument` in `pyspinw/experiment.py` to use an explicit module path.
* Added a missing import for `Path` from ~`pathlib`~  `pyspinw/path.py` in `pyspinw/measurement.py`.

### DIFF
--- a/pyspinw/measurement.py
+++ b/pyspinw/measurement.py
@@ -1,5 +1,7 @@
 """ Description of measurements """
 
+from pathlib import Path
+
 class Measurement:
     """ Provides the q and E parameters specific to a given measurement """
 
--- a/pyspinw/measurement.py
+++ b/pyspinw/measurement.py
@@ -1,6 +1,6 @@
 """ Description of measurements """
 
-from pathlib import Path
+from pyspinw.path import Path
 
 class Measurement:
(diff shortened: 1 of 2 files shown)


pyspinw/measurement.py
```diff
@@ -1,5 +1,7 @@
 """ Description of measurements """
 
+from pathlib import Path
```
mducle: I think this should be `from pyspinw.path import Path` - I'm pretty sure the path in the `OrthogonalSlice` is not a filesystem path.

mducle (changes requested):

pyspinw/measurement.py
AlbertRyu: Thanks for the info! I'm not sure I understand why the tests are failing. The only change I made was adding two imports — what could possibly break from that?

mducle:
@AlbertRyu It seems to be a random failure in one of the numerical test which could be related to the update of the runner from MacOS 15 (Sequoia) to MacOS 26 (Tahoe). I created an issue for this: https://github.com/SpinW/pySpinW/issues/203

mducle (approved):

PR #211, opened 2026-07-14
114 more viewer todos

### LINKED ISSUE #114: Viewer TODOs and proposed features

lucas-wilkins:
This checks of the following from the viewer TODO list:

- [x] Rendering of anisotropies
- [x] Toggle between perspective and orthographic views


pyspinw/gui/crystalview.py
```diff
@@ -211,6 +218,10 @@ def paintGL(self):
         spin_scale = 2 * self.display_options.atom_spin_scaling
         spin_scale_matrix = np.diag([spin_scale, spin_scale, spin_scale, 1])
 
+
+        anisotropy_scale = 2 * self.display_options.anisotropy_scaling
```
mducle: Why the hard coded factor of `2` here (and in line [218](https://github.com/SpinW/pySpinW/pull/211/changes#diff-dff5580251e4a09f948387ad244c6f17307e36d975aaff5d58a69509d49d515eR218) and `0.1` in [225](https://github.com/SpinW/pySpinW/pull/211/changes#diff-dff5580251e4a09f948387ad244c6f17307e36d975aaff5d58a69509d49d515eR225))? Maybe they can just be absorbed into the definitions in [displayoptions.py](https://github.com/SpinW/pySpinW/pull/211/changes#diff-a18913f921b2f512c4518041cc697c4907053ee9d36b245bf70964d48afe9690R23-R25)?

pyspinw/gui/rendermodel.py
```diff
@@ -125,10 +125,72 @@ class RenderAnisotropy(Component):
 
     # Don't subclass selectable as we don't want to select anisotropies from the viewer
 
-    def __init__(self, anisotropy: Anisotropy):
+    _identity = np.eye(3)
+    _max_eigen = 3**2
+    _min_eigen = 1.1**2
```
mducle: Could we change `_min_eigen` to be much smaller - e.g. `0.1**2` or `0.2**2` just so that they show up - I find that the ellipses are well, _too_ elliptical for the case of a pure axial anisotropy I'd like the figure to look much more like a straight line, and for the case of a pure planar anisotropy I'd like it to look much more like a pancake. At the moment they both look elliptical just slightly flattened or slightly pointed, so it's hard to visually diagnose the anisotropy.

pyspinw/gui/rendermodel.py
```diff
@@ -202,8 +264,27 @@ def __init__(self, hamiltonian: Hamiltonian):
...
 
         self.anisotropies = []
+
+        # We want to calculate the range of eigenvalues that we have, including zero in that
+        minimum_eigenvalue = 0.0
+        maximum_eigenvalue = 0.0
+        for anisotropy in self.expanded.anisotropies:
+            symmetrized = 0.5*(anisotropy.anisotropy_matrix + anisotropy.anisotropy_matrix.T)
+            eigenvalues = np.linalg.eigvals(symmetrized)
+            smallest = np.min(eigenvalues)
```
RichardWaiteSTFC, line 256: FYI I took the inverse of eigenvalues of anisotropy matrix to match tutorial 32? But you may be doing something different here and in `calculate_model_matrix` - so apologies if this isn't relevant!
https://github.com/SpinW/spinw/blob/dc06a4bec2c44bcfde6baa630ecb6329fb4b68ba/python/pyspinw/plotting.py#L494-L510
Btw I'm not saying what I did was rigorously correct!

pyspinw/gui/rendermodel.py
```diff
@@ -125,10 +125,72 @@ class RenderAnisotropy(Component):
...
+        #  -minimum_value / (maximum_value - minimum_value)
+        # which is the fractional distance between _min_eigen and _max_eigen.
+        #
+        # We also need to scale these to the matrices
+
+        scaling = (RenderAnisotropy._max_eigen - RenderAnisotropy._min_eigen) / (maximum_value - minimum_value)
+        base_size = RenderAnisotropy._min_eigen + -minimum_value * scaling
+        base_ellipsoid = np.diag([base_size]*3)
+
+        symmetrised = 0.5*(anisotropy.anisotropy_matrix + anisotropy.anisotropy_matrix.T)
```
mducle: There's something wrong with the ellipse calculations here - if I use `AxisMagnitudeAnisotropy` with `a>0` I would expect to see a pancake perpendicular to the `direction` whereas if I use `a<0` I would expect to see a "cigar" along the `direction` given - but at the moment it's the opposite.

This could be changed either by changing this line to use the negative of the anisotropy matrix, as in `symmetrised = -0.5*(anisotropy.anisotropy_matrix ...` and also [line 272](https://github.com/SpinW/pySpinW/pull/211/changes#diff-32f9377fd8f09e01a28dc09c2c06c13dfdd9c949bd872fe4cfa8b1dc0e6f0f45R272), or maybe to use `1 / eigvals` as Richard suggests but I couldn't get that to work here.

pyspinw/gui/rendermodel.py
```diff
@@ -125,10 +125,72 @@ class RenderAnisotropy(Component):
...
+        #  then the ellipsoid is round, at the minimum size it will have a radius
+        #  of _min_eigen and maximum size, _max_eigen
+        #
+        # The zero eigenvalue radius should then be determined by calculating:
+        #  -minimum_value / (maximum_value - minimum_value)
+        # which is the fractional distance between _min_eigen and _max_eigen.
+        #
+        # We also need to scale these to the matrices
+
+        scaling = (RenderAnisotropy._max_eigen - RenderAnisotropy._min_eigen) / (maximum_value - minimum_value)
```
mducle: This rescaling does something weird to the shapes. With the debug script If I have all the same "type" e.g. axial:

```python
    anisotropies = [
        AxisMagnitudeAnisotropy(site=x, direction=[1,0,0], a=-1),
        AxisMagnitudeAnisotropy(site=y, direction=[0,1,0], a=-1),
        AxisMagnitudeAnisotropy(site=z, direction=[0,0,1], a=-1),
        AxisMagnitudeAnisotropy(site=w, direction=[1,1,0], a=-1),
    ]
```

then it looks ok:

Image

and likewise if I set them all to planar type

```python
    anisotropies = [
        AxisMagnitudeAnisotropy(site=x, direction=[1,0,0], a=1),
        AxisMagnitudeAnisotropy(site=y, direction=[0,1,0], a=1),
        AxisMagnitudeAnisotropy(site=z, direction=[0,0,1], a=1),
        AxisMagnitudeAnisotropy(site=w, direction=[1,1,0], a=1),
    ]
```

Image

But with a mix, it just looks wrong...


```python
    anisotropies = [
        AxisMagnitudeAnisotropy(site=x, direction=[1,0,0], a=-1),
        AxisMagnitudeAnisotropy(site=y, direction=[0,1,0], a=1),
        AxisMagnitudeAnisotropy(site=z, direction=[0,0,1], a=1),
        AxisMagnitudeAnisotropy(site=w, direction=[1,1,0], a=-1),
    ]
```

Image

E.g. spin 1 (highlighted) and 4 (bottom towards us) which should be "cigars" now look like fat blobs.

mducle (changes requested):

pyspinw/gui/rendermodel.py
lucas-wilkins: I guess it depends on what you would expect to see represented, I can see how the inverse would make more sense.

lucas-wilkins: It's the difference between the ellipse representing the force of the anisotropy, and the ellipse representing where the spins like to go.

pyspinw/gui/rendermodel.py
lucas-wilkins: See my response to Richard, nothing wrong with the calculation, just I seem to have a different view of what these represent.

pyspinw/gui/rendermodel.py
lucas-wilkins: There is some change in the overall shape if you start mixing things, because of the way things get scaled to make them always come out a sensible size. It's a trade off between robustness of dealing magnitudes, and accommodating mixed systems. I was thinking that many different kinds of anisotropy would be less common, so less important.

pyspinw/gui/rendermodel.py
RichardWaiteSTFC: I see thanks, my preference (as per usual!) would be to stick with behaviour of the old MATLAB version (i.e. cigar shape would indicate easy-axis along major axis of the ellipsoid) - but I defer to @mducle!

lucas-wilkins: I think the other way to how it is makes more sense.

pyspinw/gui/rendermodel.py
lucas-wilkins: Yeah, there's some balancing to do there. 1.1 is chosen because its slightly bigger than than the atoms, so it's always drawn outside them.

lucas-wilkins: Less obvious when you do spins.

pyspinw/gui/rendermodel.py
lucas-wilkins: Something is off about the shader too.

pyspinw/gui/rendermodel.py
```diff
@@ -125,39 +125,25 @@ class RenderAnisotropy(Component):
...
-        # We want an ellipsoid such that if the eigenvectors are all zero,
-        #  then the ellipsoid is round, at the minimum size it will have a radius
-        #  of _min_eigen and maximum size, _max_eigen
-        #
-        # The zero eigenvalue radius should then be determined by calculating:
-        #  -minimum_value / (maximum_value - minimum_value)
-        # which is the fractional distance between _min_eigen and _max_eigen.
-        #
-        # We also need to scale these to the matrices
+        symmetrised = 0.5 * (anisotropy.anisotropy_matrix + anisotropy.anisotropy_matrix.T)
```
mducle: For some reason it's still giving me the opposite sense of what I expected - e.g. in `viewer_anisotropy_test_2.py` if I change all values of `a` to `-1` it gives me disks, whereas I would expect to see cigars, whilst if I change all values to `1` it gives me cigars whereas I expected disks...

But if I change this line to:

```python
        symmetrised = -0.5 * (anisotropy.anisotropy_matrix + anisotropy.anisotropy_matrix.T)
```

(change sign at front)

then it works as I expect.

mducle (changes requested):
Looks a lot better but it's still giving the opposite sense (cigars vs disks) compared to what I was expecting...

pyspinw/gui/rendermodel.py
lucas-wilkins: Done, and merging

PR #213, opened 2026-08-13
TB2J interface

mducle:
Implements a reader for TB2J output files `exchange.out` (text) and `TB2J.pickle` (binary pickle).

Example script:

```
from pyspinw import *
import matplotlib.pyplot as plt

hamiltonian = load_tb2j("exchange.out")
hamiltonian.print_summary()
path = Path([[0,0,1], [0.5,0.5,1], [0.5,0.5,0.5], [1,1,0.5], [1,1,0]])
fig = hamiltonian.spaghetti_plot(path, show=False)
plt.show()
```


pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+import ase
+import re
+from pyspinw.symmetry.unitcell import UnitCell
+from pyspinw.site import LatticeSite
+from pyspinw.exchange import Exchange, HeisenbergExchange
+from pyspinw.structures import Structure
+from pyspinw.hamiltonian import Hamiltonian
+
+EXCH_TOL = 0.1    # Tolerance below which exchange values are considered zero
```
lucas-wilkins, line 12: What units are these? Seems like a large number

pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class _coupling:
+    def __init__(self, s1, s2, icv):
+        self.s1, self.s2, self.icv = s1, s2, np.array(icv)
+    def __eq__(self, other):
+        return self.s1 == other.s1 and self.s2 == other.s2 and np.sum(np.abs(self.icv - other.icv)) < 1e-8
+    def __neg__(self):
+        return _coupling(self.s2, self.s1, -self.icv)
+
```
lucas-wilkins, line 39: Comments above here would be good, especially for the regex, and/or more explicit names

pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+                               for v in re.findall(JANI_EXPR, exch_txt, re.MULTILINE)])
+        self.noncolinear = False
+        if len(np.shape(magmoms)) > 1:
+            self.noncolinear = True
+            if self.j_iso.shape[0] != self.j_dmi.shape[0] or self.j_iso.shape[0] != self.j_ani.shape[0]:
+                raise RuntimeError("Inconsistent number of exchanges")
+        assert len(self.j_pos) == len(self.j_iso), "Inconsistent number of exchanges"
+
+    def _parse_pickle(self, data):
+        self.type, self.data = ('pickle', data)
```
lucas-wilkins, line 84: these variables are not defined in the init (=None would be the convention for this pattern)

pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+        self.struct.set_initial_magnetic_moments(self.data['magmoms'])
+        self.noncolinear = not self.data['colinear']
+        sym_index = _sym_index()
+        self.atoms = [sym_index(sym) for sym in self.struct.get_chemical_symbols()]
+        d_exch, d_dist = (self.data[k] for k in ['exchange_Jdict', 'distance_dict'])
+        ord_k = [k for k, v in d_dist.items() if k in d_exch]
+        ord_k = [ord_k[i] for i in np.argsort([v[1] for k, v in d_dist.items() if k in d_exch])]
+        self.j_pos = [[self.atoms[k[1]], self.atoms[k[2]], k[0], 1000*d_exch[k], d_dist[k][0], d_dist[k][1]]
+                      for k in ord_k]
+        self.j_iso, self.j_dmi, self.j_ani = ([1000*d_exch[k] for k in ord_k], [], [])
```
lucas-wilkins, line 95: more explict names would be good

pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def to_hamiltonian(self):
+        """ Create a SpinW Hamiltonian object from the TB2J inputs """
+        unitcell = UnitCell(*tuple(self.struct.get_cell().cellpar()))
+        # ASE and TB2J uses moments in uB; SpinW wants it as spin length S where mu = g*S
+        pos, moms = self.struct.get_scaled_positions(), self.struct.get_initial_magnetic_moments() / 2
+        # Calculates the inter-cell vector from the inter-site vector given by TB2J
+        vec = unitcell.cartesian_to_lattice_units(np.array([p[4] for p in self.j_pos]))
+        idx = {n:i for i, n in enumerate(self.atoms)}
+        ijvec = [tuple(map(int, np.floor((pos[idx[p[0]]] + dv)+0.01))) for p, dv in zip(self.j_pos, vec)]
```
lucas-wilkins, line 125: This looks a bit sketchy

pyspinw/tb2j.py
```diff
@@ -0,0 +1,165 @@
...
+        idx = [(i1,i2+i1+1) for i1, c1 in enumerate(couplings) for i2, c2 in enumerate(couplings[i1+1:]) if c1 == -c2]
+        # TB2J uses the convention that spins a normalised so a TB2J exchange = J*SiSJ - need scale this out
+        # TB2J also follows the opposite sign convention to SpinW so we have a negative here
+        mm = [np.linalg.norm(m) for m in moms]
+        jf = {f'{self.atoms[i1]}{self.atoms[i2]}':-mm[i1]*mm[i2] for i1 in range(3) for i2 in range(3)}
+        # Now construct the Hamiltonian
+        def _DM(dv):
+            return np.array([[0, dv[2], dv[1]], [-dv[2], 0, dv[0]], [-dv[1], -dv[0], 0]])
+        exchanges = []
+        if self.noncolinear:
```
lucas-wilkins, line 138: I've got a thing in the symmetry branch that will figure out the best exchange class to use, probably don't need to do any of this

pyspinw/tb2j.py
lucas-wilkins: Also type annotations would be good.

lucas-wilkins (commented):
In general, it could be a bit more readable.

pyspinw/tb2j.py
mducle: Yeah, it was made large for debugging and I forgot to set it to a sensible value. Units should be in meV.

pyspinw/tb2j.py
mducle: Added help comment.

pyspinw/tb2j.py
mducle: So this is just how TB2J stores the data - if you did a colinear spin-polarised DFT calculation it can only calculate a Heisenberg type exchange (`J_iso`). If you did a non-colinear spin-polarised DFT calculation it can calculate the off-diagonal elements (`J_ani`) and if then also include SOC (which needs a non-colinear calculation) it will also give you a DM-interaction (`DMI`). But these are given as separate values in the input file. For a colinear calculation `DMI` and `J_ani` are not written, for noncolinear calculations it will print both.

I'm just creating a general `Exchange` matrix now... we can use the symmetry machinary to work out which class is the best after.

pyspinw/tb2j.py
mducle: Suggestions welcomed!

pyspinw/tb2j.py
mducle: I just went with what TB2J called the variables...

manueldsdias:
I think that I will finally be able to read this carefully once SCARF goes on holiday. I tried the parser and noticed some flat bands in the MnF2 example. This is because both Mn and F atoms are assigned to the spin Hamiltonian but the exchange interactions are only defined between Mn atoms. To get this right there is another section of the `exchange.out` file generated by TB2J that has to be parsed, the one that starts with `Orbitals used in decomposition:`. The information about the orbitals is not needed, but it lists the atoms for which the exchange interactions are defined, e.g. `Mn1 : ['orb_1', 'orb_2', 'orb_3', 'orb_4', 'orb_5', 'orb_6']`. Grabbing `Mn1`, etc. from this would generate the list of magnetic atoms for which the interactions are actually defined. It's still helpful to read all the atoms from the `Atoms:` section of the file to be able to carry out symmetry analysis (and possibly symmetrisation of the numerical results).

pyspinw/tb2j.py
```diff
@@ -0,0 +1,187 @@
...
+        idx = {n:i for i, n in enumerate(self.atoms)}
+        ijvec = [tuple(map(int, np.floor((pos[idx[p[0]]] + dv)+0.01))) for p, dv in zip(self.j_pos, vec)]
+        # TB2J lists every pair i->j and j->i whereas SpinW implicitly calculates the reversed j->i
+        # so we need to remove these otherwise it will be double counted
+        couplings = [_coupling(p[0], p[1], ijv) for p, ijv in zip(self.j_pos, ijvec)]
+        idx = [(i1,i2+i1+1) for i1, c1 in enumerate(couplings) for i2, c2 in enumerate(couplings[i1+1:]) if c1 == -c2]
+        # TB2J uses the convention that spins a normalised so a TB2J exchange = J*SiSJ - need scale this out
+        # TB2J also follows the opposite sign convention to SpinW so we have a negative here
+        mm = [np.linalg.norm(m) for m in moms]
+        jf = {f'{self.atoms[i1]}{self.atoms[i2]}':-mm[i1]*mm[i2] for i1 in range(3) for i2 in range(3)}
```
manueldsdias, line 155: I think that here it should not be `range(3)`, it should be how many atoms there might be. Ugly but works: `range(len(mm))`.

PR #221, opened 2026-09-02
Tutorial output

### LINKED ISSUE #178: Tutorial publishing workflow
The idea for the tutorials is that they will be python files, and we make a script to convert them to html

I suggest we do the following:

* Use single hash comments for the text we want in the webpage
* Have some kind of tag for any images that need to be generated, I suggest using triple and double hash comments for the code that generates them, with the triple hash providing the information needed for including it

```python

ham.spagghetti_plot()

### generates: example.png
## plt.savefig("example.png")

```

Parsing this would be pretty simple:

* make a unique directory for each python file
* convert all single hash to text for the html
* convert all triple hash to image tags
* all non-comments go into python code snippets
* Then, to generate images:
  * uncomment all double comments
  * run script

lucas-wilkins:
Tutorial publication workflow and one and a half examples. This PR is only for the workflow, not the tutorials themselves.

 - makes an .md file from a runnable python module with the right comments (there is a doc explaining)
 - makes a .py file that matches the tutorial
 - makes a jupyter notebook that matches the tutorial
 - makes a file that generates images and captures from stdout/stderr
 - inserts images and captured text into the .md file
 - creates an `index.md` in the root directory which lists the tutorials
 - makes sure things are linked together

The ferromagnetic chain tutorial is about ready, the antiferromagnetic chain one builds, but needs work.

Currently doesn't link to the MATLAB tutorials, because we need to decide where they will live


examples/tutorials/jupyter_files.py
```diff
@@ -0,0 +1,100 @@
...
+        if not lines:
+            return
+
+        self.notebook["cells"].append(
+            {
+                "cell_type": "code",
+                "execution_count": None,
+                "metadata": {},
+                "outputs": [],
+                "source": lines
```
mducle: I think this should be `'\n'.join(lines)` - at the moment if you look at the `.ipynb` [file](https://github.com/SpinW/pySpinW/blob/af6a948b1ab1de1c07723ca8b9892489e2d526c6/examples/tutorials/tutorial_outputs/antiferromagnetic_chain/antiferromagnetic_chain.ipynb) all the code is just on one line because of the `rstrip` in line 37.

examples/tutorials/convert_tutorial_files.py
```diff
@@ -0,0 +1,341 @@
...
+
+                                                jupyter.add_text("## " + parts[1])
+
+                                                md_file.write("## " + parts[1])
+
+                                            elif line.startswith("image"): # Put an image tag in .md, and code in the python
+                                                parts = line.split(":")
+                                                image_filename = parts[1]
+                                                md_file.write(f"\n![]({image_filename})\n\n")
+                                                rest_is_code = True
```
mducle, line 186: I saw that you've got a method, `JupyterFile.add_output_image_to_last` which is not used - it wouldn't work here because the image is generated when `artifacts.py` is run but maybe the code could be restructured to write a placeholder here, and put the `jupyter.write_notebook` after the artifacts have been generated and change `write_notebook` to call `add_output_image_to_last` to replace the placeholder?

Or if we decide that we don't want any images in the generated ipynb then maybe just remove the `add_output_image_to_last` function?

examples/tutorials/jupyter_files.py
```diff
@@ -0,0 +1,100 @@
...
+            })
+
+    def add_output_image_to_last(self, image_file):
```
mducle, line 81: This function and the following `add_text_output_to_last` are not used in the script to convert tutorial files... maybe they can be removed if we decide that we don't want to pre-generate any outputs from the cells? (And rely on users running the cells to get outputs)

mducle (changes requested):
It looks good - just a couple of changes to the jupyter generation. The MD files look great.

One thing - I would rather the outputs of the tutorials *not* be included in the source - particularly the images. I think we should just keep the inputs and generate the outputs directly to the web page repository rather than keep it in the main code source.

examples/tutorials/convert_tutorial_files.py
lucas-wilkins: Yeah, I hadn't decided, was going to see what you thought. I didn't do what you were saying because I wasn't sure if it mattered, but I left the functions there in case it was important. Pretty sure they work, so no issue with keeping the object more functional.

examples/tutorials/jupyter_files.py
lucas-wilkins: yeah, decisions.

examples/tutorials/jupyter_files.py
lucas-wilkins: whoops

lucas-wilkins: Fixed

lucas-wilkins:
> One thing - I would rather the outputs of the tutorials _not_ be included in the source - particularly the images. I think we should just keep the inputs and generate the outputs directly to the web page repository rather than keep it in the main code source.

Too late, they're in there now! They'll be in the repo until the end of time, if not as files, as thier ghosts... actually, I think I remember hearing that this issue has been fixed.

lucas-wilkins:
@mducle I think if I squash all the commits on this branch, then they will go away, as long as I have deleted them.

So don't merge it.

mducle:
> Too late, they're in there now! They'll be in the repo until the end of time, if not as files, as thier ghosts... actually, I think I remember hearing that this issue has been fixed.

Yeah, everything committed to github stays there forever unless you delete the repo... 

I'm not too concerned about this particular commit since it's not so big, but if we keep all the outputs for all the tutorials that will add up, and if we change anything which affects the pictures it will add a new picture, and I'd rather not clutter the main tree with all of that.

> @mducle I think if I squash all the commits on this branch, then they will go away, as long as I have deleted them.
> 
> So don't merge it.

I think that although it won't be in the repo when you `pull` github will still have the commit, like [this one](https://github.com/SpinW/pySpinW/pull/223/changes/577bac3c53bcb3e4c64c3f1df4727db861cab9a5) which is not on the main tree

mducle (approved):
Looks good thanks.

I'm happy to keep the functions in `JupyterFile` and not create outputs in the produced `.ipynb`

